PR #29, opened 2016-05-19
Feature/user mgmt

scottnath:
Adding the user management infrastructure and routing

**Changelog**

```markdown
**New**

* user management views
* user content type creation (changed [content-types](/punchcard-cms/content-types))

**Changed**

* `machine` becomes `id` in content-types
```

---
Doesn't completely resolve a story :/


scottnath:
ignore last push. just getting latest online.

Snugug:
@scottnath yay it passed! Going to code review now!

index.js
```diff
@@ -31,6 +31,15 @@ const initApp = () => {
     app.use('/content', contentTypesRoutes);
 
     /*
+      @name users route
+      @description create routes for users
+     */
+    app.use('/users', users.userRoutes);
+
+    // TEMP: will be removed when we have a table-making process
+    users.checkUserTable();
```
Snugug: Now that we do can you update this to follow the new DB stuff?

lib/users/index.js
```diff
@@ -0,0 +1,60 @@
...
+const contentTypes = require('punchcard-content-types');
+const userRoutes = require('./routes');
+const fs = require('fs');
+const yaml = require('js-yaml');
+const knex = require('knex')(config.knex);
+
+/*
+  Checks for the existence of a `users` table
+
+  ## TODO
```
Snugug: We have this (and should be in your branch). Can you update? Add to https://github.com/punchcard-cms/punchcard/blob/master/lib/database/schemas.js and you should be GTG

lib/users/index.js
```diff
@@ -0,0 +1,60 @@
...
+ */
+const checkUserTable = () => {
+  return knex.schema.hasTable('users')
+    .then(exists => {
+      if (exists) {
+        return 'table "users" EXISTS';
+      }
+
+      return knex.schema.createTable('users', table => {
+        table.uuid('id');
```
Snugug: This doesn't need to be universally unique, only unique in the table. A simple `increments` `index` should do

lib/users/index.js
```diff
@@ -0,0 +1,60 @@
...
+        table.string('email').unique();
```
Snugug: This flag currently doesn't exist in the schema creation, but shouldn't be too hard to add

lib/users/index.js
```diff
@@ -0,0 +1,60 @@
...
+        table.string('password');
+        table.string('level');
```
Snugug: What's `level` for?

lib/users/index.js
```diff
@@ -0,0 +1,60 @@
...
+        table.jsonb('contentAccess');
```
Snugug: Can this be simplified to `access`?

lib/users/index.js
```diff
@@ -0,0 +1,60 @@
...
+        table.string('created');
```
Snugug: Should be `timestamp`

lib/users/index.js
```diff
@@ -0,0 +1,60 @@
...
+  else {
+    userConfig = yaml.safeLoad(fs.readFileSync(config.users.configFile, 'utf8'));
+  }
+
+  // return the merged content type object
+  return contentTypes([userConfig]);
+};
+
+module.exports.userContentType = userContentTypeMerged;
+module.exports.userRoutes = userRoutes;
```
Snugug: * This is the `user` module, don't need to prefix the child functions with `user`
* Whichever is the _default_ should just be `module.export`ed, the children with dot notation

Snugug: See https://github.com/punchcard-cms/content-types/blob/master/lib/content-types.js#L66-L68

Snugug:
Why isn't this being handled with [Passport](https://www.npmjs.com/package/passport)?

lib/users/routes.js
```diff
@@ -0,0 +1,85 @@
...
+    })
+    .catch(err => {
+      console.error(err);
+    });
+  });
+});
+
+// POST method route
+router.post('/add', multipartMiddleware, () => {
+  console.log('NEW user added to database because awesome');
```
Snugug: This is missing

Snugug:
If we're going to be handling passwords ourselves, need to use [Bcrypt](https://www.npmjs.com/package/bcrypt-nodejs) to secure them. No passwords in plaintext

Snugug:
https://codahale.com/how-to-safely-store-a-password/

lib/users/routes.js
```diff
@@ -0,0 +1,85 @@
...
+    return contentTypes.form(userContentType[0])
+    .then(form => {
+      res.render('users/add', {
+        pageTitle: `${config.users.usersHome.title} :: Add`,
+        formContents: form.html,
+        validation: form.validation,
+      });
+    })
+    .catch(err => {
+      console.error(err);
```
Snugug: Want to do more than `console.log` on error

lib/users/routes.js
```diff
@@ -0,0 +1,85 @@
...
+
+// POST method route
+router.post('/add', multipartMiddleware, () => {
+  console.log('NEW user added to database because awesome');
+});
+
+
+/*
+User edit entry page
+TODO: inject content into form
```
Snugug: Let's actually do this please

lib/users/index.js
scottnath: role: superuser/admin/creator

tests/users.js
```diff
@@ -0,0 +1,46 @@
...
+
+test('Exports exist', t => {
+  t.is(typeof users, 'object', 'Users exports a function');
+
+  t.is(typeof users.userContentType, 'function', 'Submodule `userContentType` exists and is a function');
+  t.is(typeof users.userRoutes, 'function', 'Submodule `userRoutes` exists and is a function');
+  t.is(typeof users.checkUserTable, 'function', 'Submodule `checkUserTable` exists and is a function');
+});
+
+test('Users, with config, merged with correct param', t => {
```
Snugug: Don't test your dependencies. There should be tests inside of `content-types` to test this functionality, you don't need to.

Snugug: And should not

lib/users/index.js
```diff
@@ -0,0 +1,60 @@
...
+  }
+  else {
+    userConfig = yaml.safeLoad(fs.readFileSync(config.users.configFile, 'utf8'));
+  }
+
+  // return the merged content type object
+  return contentTypes([userConfig]);
+};
+
+module.exports.userContentType = userContentTypeMerged;
```
Snugug: This isn't a content type. Don't call it a content type.

lib/users/index.js
Snugug: `routes` should suffice

Snugug:
Keep rolling on this @scottnath. When it's been updated ping me please

lib/users/routes.js
scottnath: we still don't have a plan on this. See [this story on error handling](https://github.ibm.com/Watson/holmes/issues/101). If we have a temporary solution we're using, glad to use it.

lib/users/index.js
scottnath: I found this as part of KNEX.

lib/users/routes.js
scottnath: @Snugug how do I get the data from the form? This is what `req.body` is returning:

```
{ '362978f8-a24d-4e09-8842-bb017da032d5': 'email@test.com',
  '4e95dbbd-10de-4da5-8b60-dfba3f127279': '111222',
  '2b1352c4-220c-4187-887f-7cc62cdb16ee': 'superuser' }
```

Is there a way for me to translate those ids into which input they represent in the current punchcard?

Snugug: Yup, cross-check with the merged result

Snugug: I think I'm going to build in verification in to Content Types

lib/users/routes.js
scottnath: Same question as the response from the user form. How do I propagate the form if I don't know the `name` field for the input type?

Snugug: You'll know the IDs, although the name stuff may need to be updated in `content-types`

lib/users/routes.js
Snugug: The standard way is include `next` in your func params and on error, call `next(err)`

Snugug: Let's do that for now until we come up with a more universal way forward

lib/users/index.js
```diff
@@ -0,0 +1,37 @@
+/**
```
Snugug: The convention we have in place is to have `lib/thing.js` and `lib/thing/foo.js`, not `lib/thing/index.js`. Please either follow the existing convention, or update the other files to this convention. FWIW this is easier to tab to in the command line

lib/users/index.js
```diff
@@ -0,0 +1,37 @@
...
+  else {
+    userConfig = yaml.safeLoad(fs.readFileSync(config.users.configFile, 'utf8'));
+  }
+
+  // return the merged content type object
+  return contentTypes([userConfig]);
+};
+
+module.exports = userModelMerged;
+module.exports.userRoutes = userRoutes;
```
Snugug: Should be `routes` not userRoutes

index.js
```diff
@@ -31,6 +31,12 @@ const initApp = () => {
     app.use('/content', contentTypesRoutes);
 
     /*
+      @name users route
+      @description create routes for users
+     */
+    app.use('/users', users);
```
Snugug: Does this work? The default function for `users` is the merged user model, not the routes

lib/users/index.js
```diff
@@ -0,0 +1,37 @@
...
+const userRoutes = require('./routes');
+
+/*
+ * Returns a content model object with merged input plugin data
+ *
+ * @param {object} userContentType  data on a content type and the plugins it wants to use
+ *
+ * @returns {object} data on a content type merged with input plugin data
+ */
+const userModelMerged = (userContentType) => {
```
Snugug: We are in the User file, we don't need to prefix everything with `user`

Issue #19, opened 2016-05-16
Core Developer: Database schema


scottnath:
_As a_ Core Developer _I want_ database schema created from content types _so that_ the database storage schema matches the requirements of our data model.

```gherkin
Feature: Database schema creation
  As a Core Developer
  I want database schema created from content types
  So that the database storage schema matches the requirements of our data model.

  Scenario: use content model to create database schema
    Given that there is data within a content model which directly relates to setting up database tables
    When the system is creating or updating database schema
    Then it should create the schema from each content model's attributes.

  Scenario: cross-content type attributes for database schema
    Given that there are attributes which are common across content types (such as sunrise/sunset/id/created/etc/etc)
      And that these attributes don't belong in a content model
    When I am creating a content-type's database schema
    Then I need a way to use these common attributes, along with the attributes within the content model, when creating database schema.

Feature: Update Schema in Database
  As a Core Developer
  I want to update the schema in the database
  So that the database schema is ready for our content's specific structure.

  Scenario: use database schema to make new tables in the db
    Given that we want to automate the process of creating database tables
      And that we want the tables to match *exactly* the data types of our content model
    When I add a new content model to the CMS
    Then the system should determine the database schema for that model
      And it should use that schema to create a new table in the database.

  Scenario: use database schema to update existing tables in the db
    Given that we want to automate the process of updating database tables when inevitable changes to the content model happen
      And that we want the tables to match *exactly* the data types of our content model
    When I changing new content model in the CMS
    Then the system should determine the database schema for that model
      And it should use that schema to update the corresponding table in the database
      And it should not lose any existing data in the process.
```

---

- [ ] Schema creation / use content model
- [ ] Schema creation / cross-content attributes
- [ ] Update schema / new tables
- [ ] Update schema / existing tables

Snugug:
Schema

![Schema](https://cloud.githubusercontent.com/assets/377188/15398110/6796bffa-1db0-11e6-935e-489a98fa8ee2.JPG)

PR #30, opened 2016-05-20
Database Schema

Snugug:
Add Database Schema creation, and refactor the _crap_ outta everything to support that

**Changelog**

```markdown
**New**

* Database Schema Creation on Initialization 

**Changed**

* Index now returns a promise. :boom:

**Removed**

* Remove old tests
* Remove old schema
```

---
Resolves: #19 

`DCO 1.1 Signed-off-by: Sam Richard <snugug@us.ibm.com>`


lib/database.js
```diff
@@ -0,0 +1,57 @@
...
+
+const tables = require('./database/tables');
+const schemas = require('./database/schemas');
+const utils = require('./utils');
+
+const knex = require('knex');
+
+const database = knex(config.knex);
+
+const init = () => {
```
scottnath: add jsdoc please

lib/database/schemas.js
```diff
@@ -0,0 +1,75 @@
...
+  {
+    name: 'created',
+    type: 'timestamp',
+  },
+  {
+    name: 'language',
+    type: 'string',
+  },
+  {
+    name: 'sunrise',
```
scottnath: At one point you said `sunrise` would be `required`. Still so?

lib/database/schemas.js
```diff
@@ -0,0 +1,75 @@
...
+    name: 'id',
+    type: 'string',
+  },
+  {
+    name: 'created',
+    type: 'timestamp',
+  },
+  {
+    name: 'language',
+    type: 'string',
```
scottnath: Language should be a `select` or point to a taxonomy or some such.
Does _not_ have to be for this pull request...but will need to happen

lib/database/schemas.js
scottnath: If `required` doesn't belong here...where does it belong?

lib/database/schemas.js
Snugug: This is Database schema, _not_ Content Model schema

lib/database/schemas.js
Snugug: This is database schema, _not_ Content Model schema

scottnath: Cool....where will these common content model schemas live?

Snugug: Probably inside `lib/content-types`. They aren't schema. The user mocking will live inside `lib/auth` or `lib/user` or something like that.

lib/database.js
```diff
@@ -0,0 +1,57 @@
...
+    // Loop over all schema and create database tables
+    return Promise.map(schemata, schema => {
+      return tables.create(database, schema);
+    });
+  }).then(() => {
+    // Get raw content types
+    return types.raw();
+  }).then(raw => {
+    // Store versioned content types in DB
+    return database('all-types').insert({
```
scottnath: Can this please move this to initialization.

scottnath:
Just those notes. Jsdocs missing throughout, but can be fixed. Only real item is moving `raw` out of database initialization.
:hammer:

scottnath:
@Snugug :/ failing test

Snugug:
I fixed the damn race condition

Issue #9, opened 2016-05-16
Administrator: User Management & Levels


scottnath:
_As a_ Administrator _I want_ to manage users _so that_ I can control content access, roles, permissions, sessions, and logins for each user in the system.

```gherkin
Feature: Manage Users
  As an Administrator
  I want to manage CMS users
  So that I can control content access, roles, permissions, sessions, and logins for each user in the system.

  Scenario: Add a user
    Given that I need to give new users access to the CMS
    When I want to add a user
    Then there should be functionality to add a new user
      And this functionality should be limited to Administrators.

  Scenario: Edit a user
    Given that I may need to edit a user's login credentials
    When I want to update a user's information
    Then there should be functionality to edit a user
      And this functionality should be limited to Administrators.

  Scenario: Delete a user
    Given that I may need to delete a user from access to the CMS
    When I want to delete a user
    Then there should be functionality to delete a user
      And this functionality should be limited to Administrators.

  Scenario: Change a user's password
    Given that I may need to change a user's password
    When I want to update a user's password
    Then there should be functionality to change a password
      And this functionality should be limited to that user and Administrators.
 
  Scenario: Select user access level
    Given that the CMS will have require different levels of access
    When I want to give users access at different levels
    Then there should be a way to select an access level for a user

  Scenario: Limited user levels
    Given that the CMS needs specific user levels
    When I select a level from the user levels available
    Then one level should allow full access all aspects of the system
      And one level should allow content access only
      And one level should allow specific content-type access only

Feature: User management dashboard
  As an Administrator
  I want to have a central location for managing all users
  So that I can find and manage individual users or add new users.

  Scenario: See all users
    Given that I need to see all users in the system
    When I attempt to manage users
    Then I need an interface that lists all users

  Scenario: User actions
    Given that I need to edit or delete a user
    When viewing users in the user-mgmt interface
    Then each user should have a CTA for editing the user
      And each user should have a CTA for deleting the user.

  Scenario: Add a user CTA
    Given that I will have a user management interface
    When I want to add a new user
    Then there should be a CTA to go to an interface to add a new user.
```

---

- [x] Manage Users / add
- [x] Manage Users / edit
- [x] Manage Users / delete
- [ ] Manage Users / password change
- [x] Manage Users / select access level
- [ ] Manage Users / access levels available
- [x] Dashboard / see users
- [x] Dashboard / user actions
- [x] Dashboard / add user cta

scottnath:
[...]

Issue #8, opened 2016-05-16
Content Creator: user login functionality


scottnath:
_As a_ content creator _I want_ have login functionality _so that_ I can log in to the site and control my login credentials.

```gherkin
Feature: User Login functionality
  As a Content Creator
  I want to have login functionality
  So that I can log in to the site and control my login credentials.

  Scenario: Login required to view CMS
    Given use of the CMS is restricted
    When attempting to view pages in the CMS
    Then I must be required to log in before seeing pages.

  Scenario: Login requires a username and password
    Given that I must have unique credentials to log in to the CMS
    When logging into the CMS
    Then I must have a unique username
      And I must have a secure password.

  Scenario: Auth State
    Given that I must be authenticated to view the CMS
      And that this authentication is given to me after a successful login
    When I log in in to the CMS
    Then the authorization I receive must be maintained while I'm using the CMS
      And maintain my ability to use the CMS.

---

Feature: Technical functionality
  As a Core Developer
  I want to integrate login functionality into Punchcard
  So that authentication of users limits access to Punchcard
    And so someone gives me a Cheers for Peers.

  Scenario: Username and password storage
    Given that usernames and passwords must be secured
    When the systems needs to store or retrieve a user's credentials
    Then those credentials should be in the database.
```

---

- [x] User Login functionality / login required
- [x] User Login functionality / username&password
- [x] User Login functionality / auth state
- [x] Technical / u&p storage

scottnath:
User Login functionality / password recovery & change password moved to #148 

closing this issue as all other work complete.

PR #37, opened 2016-05-25
Feature/user auth

scottnath:
User authentication and management

**Changelog**

```markdown
**New**

* login authentication
* roles
* bcrypt passwords
* edit users
* add users
* delete users

**Changed**

* dep upgrades
```

---
Resolves: #8
Resolves: #9 


Snugug:
@scottnath Your coverage dropped too much. Please increase your code coverage.

scottnath:
@snugug agree and aware. Can you please go over the pr anyway (as time
allows) so I can tackle en masse mañana?

config/default.js
```diff
@@ -29,6 +29,10 @@ module.exports = {
   site: {
     name: 'Punchcard CMS',
   },
+  login: {
+    loginPath: '/login',
```
Snugug: Can we work on naming things? Let's try and have as few camelCase names as we can. I'd prefer `users.paths.login` to `login.loginPath`

lib/auth/login.js
```diff
@@ -0,0 +1,65 @@
...
+ * @property {string} role - User role level
+ */
+
+/*
+ * Check login status
+ *
+ * @param {ensureLoggedIn--options} opt options for login check
+ */
+const ensureLoggedIn = (opt) => {
+  let options;
```
Snugug: const before let plz

Snugug: And please initialize immediately. `let options = opt`

lib/auth/login.js
```diff
@@ -0,0 +1,65 @@
...
+
+  if (!options.hasOwnProperty('redirectTo')) {
+    options.redirectTo = config.login.loginPath;
+  }
+
+  return (req, res, next) => {
+    // check if authenticated
+    if (!req.isAuthenticated || !req.isAuthenticated()) {
+      if (req.session) {
+        req.session.returnTo = req.originalUrl || req.url; // eslint-disable-line no-param-reassign
```
Snugug: If you're going to disable a linter, please add a comment as to why

lib/auth/login.js
```diff
@@ -0,0 +1,65 @@
...
+      }
+
+      return res.redirect(url);  // eslint-disable-line newline-before-return
```
Snugug: Why did you disable this? Especially because you appear to be doing it?

lib/auth/login.js
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    // check for role
+    if (options.hasOwnProperty('role')) {
+      if (req.user.role !== options.role) { // TODO: needs to be changed to check level is WITHIN range
+        const err = new Error('Permission denied: you do not have access to this cool page.');
```
Snugug: This error message we're going to want to be i18n friendly, so we don't want to hard-code it.

lib/auth/login.js
```diff
@@ -0,0 +1,65 @@
...
+        res.render('error', {
+          message: err.message,
+          error: err,
+        });
+      }
+    }
+
+    next();
+
+    return true;
```
Snugug: What's this return for?

lib/auth/login.js
```diff
@@ -0,0 +1,65 @@
...
+ * @param {ensureLoggedIn--options} opt options for login check
+ */
+const ensureLoggedIn = (opt) => {
+  let options;
+  const url = config.login.loginPath;
+  options = opt;
+
+  if (typeof options !== 'object') {
+    options = {
+      redirectTo: config.login.loginPath,
```
Snugug: Again, naming things.

`options.redirect` and `options.return`

lib/init/passport.js
```diff
@@ -0,0 +1,65 @@
...
+  {
+    usernameField: 'email',
+  },
+  (username, password, cb) => {
+    database.select('*').from('users').where({
+      email: username,
+    })
+    .then(user => {
+      if (user.length < 1) {
+        return cb(null, false, { message: 'Email not in system' });
```
Snugug: Don't hardcode user facing messages

lib/init/passport.js
```diff
@@ -0,0 +1,65 @@
...
+      }
+
+      if (!bcrypt.compareSync(password, user[0].password)) {
+        return cb(null, false, { message: 'Incorrect password' });
```
Snugug: Don't hardcode user facing messages

lib/init/passport.js
```diff
@@ -0,0 +1,65 @@
...
+      }
```
Snugug: From a [security perspective](http://ux.stackexchange.com/questions/13516/how-to-tell-the-user-his-login-credentials-are-incorrect) incorrect username and password should have the same messaging. While telling a user that either the email address or password is incorrect, if someone wanted to be malicious, they could keep guessing combos until they got an "incorrect password" error, then they would know the email exists in the system.

lib/routes/index.js
```diff
@@ -5,14 +5,64 @@
...
 
 /*
 Home Page Router
 @function
 @name /
 */
-router.get('/', (req, res) => {
-  res.render('index', { pageTitle: 'Punchcard Home' });
+router.get('/',
+  loginMiddleware(),
```
Snugug: Does putting it just on `/` cover _all_ routes or _just_ `/`? Like if I wanted to go to `/content` would that be covered by the login middleware?

lib/routes/index.js
```diff
@@ -5,14 +5,64 @@
...
+      error: err,
+    });
+  });
+
+/*
+Login Page Router
+@function
+@name /login
+*/
+router.get('/login', (req, res) => {
```
Snugug: You've got config for this route but don't use it here

lib/routes/index.js
```diff
@@ -5,14 +5,64 @@
...
+    pageTitle: 'Login',
+  });
+});
+
+/*
+Logout Router
+@function
+@name /logout
+*/
+router.get('/logout', (req, res) => {
```
Snugug: Config for this

lib/routes/index.js
```diff
@@ -5,14 +5,64 @@
...
+  req.logout();
+  res.redirect('/login');
 });
 
+// POST route
+router.post('/login', passport.authenticate('local', {
```
Snugug: Config for thsi

lib/routes/index.js
Snugug: Login/Logout should have their own routes file, not throw in to `index.js`

lib/users/data.js
```diff
@@ -31,4 +32,27 @@ const modelMerged = (userModel) => {
...
+      value = bcrypt.hashSync(value);
+    }
+    cleanData[newKey] = value;
+  });
+
+  return cleanData;
+};
+
 module.exports = modelMerged;
+module.exports.cleanFormData = cleanFormData;
```
Snugug: soMuchCamelCase. Can you come up with names that have fewer words

Snugug: youSoundLikeAJavaDeveloper

lib/users/data.js
```diff
@@ -31,4 +32,27 @@ const modelMerged = (userModel) => {
...
 };
 
+/*
+ * Cleaner user data post form submission
+ *
+ * @param {object} body req.body post-form submission
+ *
+ * @returns {object} cleaned version of form data
+ */
+const cleanFormData = (body) => {
```
Snugug: Why does this need to exist? Why can't you just use the form field names? And why is the bcrypt hash part of cleaning up the form data?

lib/users/routes.js
```diff
@@ -10,58 +10,79 @@ const contentTypes = require('punchcard-content-types');
 const multipart = require('connect-multiparty');
```
Snugug: Why isn't this `lib/routes/user.js`?

lib/users/routes.js
```diff
@@ -10,58 +10,79 @@ const contentTypes = require('punchcard-content-types');
...
+  (req, res, next) => {
+    let content;
+    database.select('*').from('users')
+    .then(rows => {
+      content = rows;
+    })
+    .catch(error => {
+      next(error);
+    })
+    .finally(() => {
```
Snugug: Finally runs even if `error` happens. This should only happen on a success

lib/users/routes.js
```diff
@@ -10,58 +10,79 @@ const contentTypes = require('punchcard-content-types');
...
+      res.render('users/landing', {
+        pageTitle: config.users.usersHome.title,
```
Snugug: title

lib/users/routes.js
```diff
@@ -10,58 +10,79 @@ const contentTypes = require('punchcard-content-types');
...
+        usersContent: content,
```
Snugug: users.content

lib/users/routes.js
```diff
@@ -10,58 +10,79 @@ const contentTypes = require('punchcard-content-types');
...
+        usersConfig: config.users,
```
Snugug: users.config

lib/users/routes.js
```diff
@@ -10,58 +10,79 @@ const contentTypes = require('punchcard-content-types');
...
+        user: req.user,
```
Snugug: users.active

lib/users/routes.js
```diff
@@ -10,58 +10,79 @@ const contentTypes = require('punchcard-content-types');
...
+    'redirectTo': '/login',
+    'setReturnTo': '/users/add',
+  }),
+  (req, res, next) => {
+    users().then(userModel => {
+      contentTypes.form(userModel[0])
+      .then(form => {
+        res.render('users/add', {
+          pageTitle: `${config.users.usersHome.title} :: Add`,
+          formContents: form.html,
```
Snugug: Just `form`, then from your template, `form.html` and `form.validation`

lib/users/routes.js
```diff
@@ -71,16 +92,126 @@ TODO: inject content into form
...
+  loginMiddleware({
+    'role': 'superadmin',
+    'redirectTo': '/login',
+    'setReturnTo': '/users/add',
+  }),
+  (req, res, next) => {
+    req.session.editUserId = req.params.id; // eslint-disable-line no-param-reassign
+
+    database('users').where('id', req.params.id)
+    .then(rows => {
```
Snugug: Please be consistent with where you put the first `.then`. It should go after the call, not on the next line

lib/users/routes.js
```diff
@@ -71,16 +92,126 @@ TODO: inject content into form
...
-  res.render('users/add', {
-    pageTitle: `${config.users.usersHome.title} :: Edit User`,
+router.get('/edit/:id',
+  loginMiddleware({
+    'role': 'superadmin',
+    'redirectTo': '/login',
+    'setReturnTo': '/users/add',
+  }),
+  (req, res, next) => {
+    req.session.editUserId = req.params.id; // eslint-disable-line no-param-reassign
```
Snugug: allTheCamelCase. `req.session.form.id` maybe. Something like that, so we can have a consistent place to throw sessions stuff for active form

lib/users/routes.js
```diff
@@ -71,16 +92,126 @@ TODO: inject content into form
...
+
+    database('users').where('id', req.params.id)
+    .then(rows => {
+      if (rows.length < 1) {
+        const err = new Error(`User with id ${req.params.id} does not exist`);
```
Snugug: Don't hard code

lib/users/routes.js
```diff
@@ -71,16 +92,126 @@ TODO: inject content into form
...
+        err.status = 404;
+        res.render('error', {
+          message: err.message,
+          error: err,
+        });
+      }
+
+      users().then(userModel => {
```
Snugug: `user` or `model`, not `userModel`

lib/users/routes.js
```diff
@@ -71,16 +92,126 @@ TODO: inject content into form
...
+        // remove passwoord
+        // TODO: password manipulation
+        userModel[0].attributes.splice(1, 1);
```
Snugug: What is this? What is this doing?

lib/users/routes.js
```diff
@@ -71,16 +92,126 @@ TODO: inject content into form
...
+
```
Snugug: No error checking for multiple rows.

Either check for multiple and throw an error, or save the single item for reuse instead of doing `rows[0]` everywhere

lib/users/routes.js
```diff
@@ -71,16 +92,126 @@ TODO: inject content into form
...
+          'role': {
+            'select': {
+              'value': rows[0].role,
+            },
+          },
+        }, userModel)
+        .then(result => {
+          contentTypes.form(result).then(form => {
+            res.render('users/add', {
+              pageTitle: `${config.users.usersHome.title} :: Edit :: ${req.params.id}`,
```
Snugug: Don't hard code

lib/users/routes.js
```diff
@@ -71,16 +92,126 @@ TODO: inject content into form
...
+              formContents: form.html,
+              validation: form.validation,
+              user: req.user,
+            });
+          });
+        })
+        .catch(error => {
```
Snugug: If you return your promises (which you should be doing and aren't) the errors will bubble and you'll only need one `catch`

lib/users/routes.js
```diff
@@ -71,16 +92,126 @@ TODO: inject content into form
...
+    });
   });
+
+// POST method route
+router.post('/edit/:id', multipartMiddleware, (req, res) => {
+  const cleanData = users.cleanFormData(req.body);
+
+  database('users').where('id', '=', req.session.editUserId).update(cleanData)
+    .then(() => {
+      req.session.editUserId = ''; // eslint-disable-line no-param-reassign
```
Snugug: Why are you throwing this in to sessions on a successful insert? What does that matter?

lib/users/routes.js
```diff
@@ -71,16 +92,126 @@ TODO: inject content into form
...
+@name /users/delete/{{id}}
+*/
+router.get('/delete/:id',
+  loginMiddleware({
+    'role': 'superadmin',
+    'redirectTo': '/login',
+    'setReturnTo': '/users/add',
+  }),
+  (req, res, next) => {
+    req.session.deleteUserId = req.params.id; // eslint-disable-line no-param-reassign
```
Snugug: Inside same object as where you're moving `editUserId`

lib/users/routes.js
```diff
@@ -71,16 +92,126 @@ TODO: inject content into form
...
+
+    database('users').where('id', req.params.id)
+    .then(rows => {
+      if (rows.length < 1) {
+        const err = new Error(`User with id ${req.params.id} does not exist`);
```
Snugug: Don't hard code

lib/users/routes.js
```diff
@@ -71,16 +92,126 @@ TODO: inject content into form
...
+  });
+
+
 // POST method route
-router.post('/edit/:id', multipartMiddleware, () => {
-  console.log('waiting on content types form parsing');
+router.post('/delete/:id', multipartMiddleware, (req, res) => {
+  database('users').where('id', '=', req.session.deleteUserId).del()
+    .then(() => {
+      req.session.deleteUserId = ''; // eslint-disable-line no-param-reassign
```
Snugug: Why?

lib/users/users.yml
```diff
@@ -10,7 +10,18 @@ attributes:
     id: password
     name: Password
     description: Make it secure!
-  - type: text
-    id: level
-    name: Access Level
-    description: (make multi-select plugin w/selection of CTs for `content-creator`) CHOOSE one - superuser/admin/content-creator
+  - type: select
```
Snugug: Roles need to come from user config, not hard-coded in to Punchcard.

lib/users/users.yml
```diff
@@ -10,7 +10,18 @@ attributes:
     id: password
     name: Password
     description: Make it secure!
```
Snugug: Needs to come from config

tests/server.js
```diff
@@ -12,7 +12,7 @@ test.beforeEach((t) => {
   });
 });
 
-test.cb('CMS Landing Page', t => {
+test.cb.skip('CMS Landing Page', t => {
```
Snugug: Why are you skipping this test?

Snugug:
Lots of comments here. Many apply multiple times, but didn't mark them multiple times. Please address and make sure you don't drop in code coverage.

lib/auth/login.js
scottnath: That's why I disabled it. If I didn't, it gave me the error even with the return on line 43.

lib/users/data.js
scottnath: ☕

lib/users/data.js
scottnath: ## names
If I used the form-field names, that would mean the schema in the database would have to be written to match the form (`email--email` instead of `email`; `role--select` instead of `role`)

## bcrypt
Where should this happen?

Snugug: Maybe the problem is this is really poorly named? And that it feels like an extra step?

You could very easily do:

```javascript
db('users').insert({
  email: req.body['email--email'],
  role: req.body['email--role'],
  password: bcrypt.hashSync(req.body['email--password']),
});
```

Without this abstraction that's quite a bit confusing

lib/users/routes.js
scottnath: `// remove passwoord`
because passwoord manipulation isn't working yet, but I still need the edit form hooked up. When I get passwoord manipulation working I'll remove this. Thus the comments.

lib/users/routes.js
scottnath: I was trying to remove from sessions. Should be 
`delete reg.session.edit.user.id`

lib/users/users.yml
scottnath: This was meant to be the user config. This .yml file is being sent through the `content-types` module to create forms.
1. should I put this file somewhere else?
2. should I split out roles and pull them in to this file? Not sure how to do that with yaml, so I'd need to program for that (no biggie, just want to do it best)

tests/server.js
scottnath: Because, for now, I did not know how to add logins to test a logged in state, but did not want to comment out the test so that the `skip` would show up as a reminder to revisit. And we are very short on ⌚

scottnath:
@Snugug great comments. Only question (for now) is where should I put all that messaging? Make a new object in `config`, yes? Separate messaging file?

Snugug:
How does this relate to #44

PR #44, opened 2016-05-27
Users

scottnath:
@Snugug not an official PR. Please offer feedback on this code so I can make changes on tuesday.


config/all/roles.js
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * @returns {array} Array of user roles
+ */
+module.exports = [
+  {
+    name: 'Administrator',
+    id: 'admin',
+    allows: [
+      {
+        resources: '/content/services',
```
Snugug: Will we need to do this for every single content type?

config/default.js
```diff
@@ -44,18 +44,50 @@ module.exports = {
...
       },
+      {
+        path: 'users',
+        label: 'Users',
+      },
     ],
   },
+  authentication: {
+    messages: {
+      badlogin: 'Email and/or passowrd incorrect',
```
Snugug: `login.error` instead of `badLogin`

config/default.js
```diff
@@ -44,18 +44,50 @@ module.exports = {
...
+    },
+  },
   env,
   cookies: {
     secure: false,
     secret: process.env.COOKIE_SECRET || 'babka',
   },
+  roles: {
+    config: {
+      file: path.join(__dirname, 'all/roles.js'),
```
Snugug: What? Really? Why not `require` it?

lib/auth/acl.js
```diff
@@ -0,0 +1,80 @@
+'use strict';
+
+/**
+ * @fileoverview Access Control Lists
+ */
+// config is used below, eslint
+const config = require('config'); // eslint-disable-line no-unused-vars
```
Snugug: Why is this disabled?

lib/auth/acl.js
```diff
@@ -0,0 +1,80 @@
...
+const NodeAcl = require('acl');
+
+const userRoles = require('../../config/all/roles');
```
Snugug: Def don't do this. This should be coming from config.

Snugug: And how about `roles`

Snugug: n/m

config/all/roles.js
Snugug: Can we abstract this for ease of our users? Instead of exposing ACL directly to them?

lib/auth/acl.js
```diff
@@ -0,0 +1,80 @@
...
+  if (rolesConfig && Array.isArray(rolesConfig)) {
+    aclRoles = rolesConfig.map((role) => {
+      return {
+        roles: role.id,
+        allows: role.allows,
+      };
+    });
+  }
+
+  return aclRoles;
```
Snugug: This will return `undefined` if `rolesConfig` is not an array or is undefined. Do we want to be returning `undefined`? if we want to return a null value, it should be `null`, otherwise it should be an empty array.

config/all/roles.js
scottnath: @Snugug There are other ways to do this, but was getting very heavy for this PR. Ok if I make a story for that?

lib/auth/acl.js
scottnath: because it is used below (see comment, line 6) and eslint is not seeing it.

lib/auth/acl.js
```diff
@@ -0,0 +1,80 @@
...
+  if (req.user) {
+    acl.addUserRoles(getUserId(req), req.user.role, (err) => {
+      if (err) throw new Error(err);
+    });
+  }
+  next();
+};
+
+// Define roles, resources and permissions
+acl.allow(roles());
```
Snugug: I don't understand this. You define the function up top, with config parameter, then call it before exporting anything meaning it gets run on first initialization, meaning no change to actually pass the config.

lib/init/parser.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+const parser = require('body-parser');
```
Snugug: See if you can replace this with `connect-multiparty` so we don't have two body parsers

lib/init/passport.js
```diff
@@ -0,0 +1,68 @@
+'use strict';
+
+/**
+ * @fileoverview Passport authentication init
+ */
+const config = require('config');
+const passport = require('passport');
+const LocalStrategy = require('passport-local').Strategy;
```
Snugug: We're eventually going to want to expose the strategies in a configurable way, but this is OK for now

lib/init/passport.js
```diff
@@ -0,0 +1,68 @@
...
+const bcrypt = require('bcrypt-nodejs');
+
+const acl = require('../auth/acl');
+
+const database = require('../database');
+
+// Passport Authentication setup
+passport.use(new LocalStrategy(
+  {
```
Snugug: God this is odd. We don't follow this pattern anywhere else. I don't know what to do, though. Maybe `LocalStrategy({ usernameField: 'email' }, (username, password, cb) =>`

lib/init/passport.js
```diff
@@ -0,0 +1,68 @@
...
+    database.select('*').from('users').where({
+      email: username,
+    })
+    .then(user => {
+      if ((user.length < 1) || (!bcrypt.compareSync(password, user[0].password))) {
+        return cb(null, false, { message: config.authentication.messages.badlogin });
+      }
+
+      // register user's role for access
+      acl.addUserRoles(user[0].id.toString(), user[0].role);
```
Snugug: `addUserRoles`? How about `addRoles` or `roles`? Or `register` as we've got `reregister` to re-register a user's roles

lib/init/passport.js
```diff
@@ -0,0 +1,68 @@
...
+const passportInit = (app) => {
+  return new Promise((res) => {
+    app.use(passport.initialize());
+    app.use(passport.session());
+
+    res(app);
+  });
+};
+
+module.exports = passportInit;
```
Snugug: There's a lot of stuff in here. Don't know if that's good or bad, just something to think about.

lib/routes/login.js
```diff
@@ -0,0 +1,57 @@
...
+
+    /*
+     * @name Login Page Router
+     *
+     * @param {object} req - HTTP Request
+     * @param {object} res - HTTP Response
+    */
+    app.get('/login', (req, res) => {
+      res.render('login', {
+        pageTitle: 'Login',
```
Snugug: `title` man, `title`.
AND THROUGH CONFIG for i18n cause 😝

lib/routes/login.js
```diff
@@ -0,0 +1,57 @@
...
+  return new Promise(resolve => {
+    const app = application;
+
+    /*
+     * @name Login Page Router
+     *
+     * @param {object} req - HTTP Request
+     * @param {object} res - HTTP Response
+    */
+    app.get('/login', (req, res) => {
```
Snugug: Config please

lib/routes/login.js
```diff
@@ -0,0 +1,57 @@
...
+      });
+    });
+
+    /*
+     * @name Logout Page Router
+     *
+     * @param {object} req - HTTP Request
+     * @param {object} res - HTTP Response
+    */
+    app.get('/logout', (req, res) => {
```
Snugug: Config please

lib/routes/login.js
```diff
@@ -0,0 +1,57 @@
...
+      req.logout();
+      res.redirect('/login');
```
Snugug: Config please

lib/routes/users.js
```diff
@@ -0,0 +1,225 @@
...
+ * @fileoverview Login routing
+ */
+const config = require('config');
+const types = require('punchcard-content-types');
+const multipart = require('connect-multiparty');
+const multipartMiddleware = multipart();
+const bcrypt = require('bcrypt-nodejs');
+
+const acl = require('../auth/acl');
+const getUserId = require('../auth/acl').getUserId;
```
Snugug: Why do this instead of `acl.getUserId`? And can you use a better name, like `user`?

config/all/roles.js
scottnath: @Snugug https://github.com/punchcard-cms/punchcard/issues/47

lib/auth/acl.js
scottnath: it does. it uses the roles config on init and sets up the roles.
https://github.com/OptimalBits/node_acl#allow

lib/init/passport.js
scottnath: not mine to rename

https://github.com/OptimalBits/node_acl#addUserRoles

lib/auth/acl.js
scottnath: made it default to `[]`

lib/init/parser.js
scottnath: This is internal to Passport. Remember the bug last friday? Yeah...that's why this is here.

Without it, passport silently, painfully silently, fails.

scottnath:
@Snugug all PR changes are in. I've very close on the tests, but it just won't recognize the cookies.

config/default.js
```diff
@@ -44,18 +45,58 @@ module.exports = {
         path: 'content',
         label: 'Content',
       },
+      {
+        path: 'users',
```
Snugug: Please follow the pattern in [content](https://github.com/scottnath/punchcard/blob/80084ed4e48f29117256ca3f819b1b23eaf4c2da/config/default.js#L17-L20)

config/default.js
```diff
@@ -44,18 +45,58 @@ module.exports = {
...
+      },
     ],
   },
+  authentication: {
+    messages: {
+      login: {
+        error: 'Email and/or passowrd incorrect',
+      },
+    },
+    login: {
```
Snugug: Content pattern

config/default.js
```diff
@@ -44,18 +45,58 @@ module.exports = {
...
+    config: roles,
+  },
   users: {
-    usersHome: {
+    home: {
       path: '/users',
       title: 'Users',
-      desc: 'This is the users landing page.',
+    },
+    add: {
```
Snugug: Can be created via `base` and `action`

lib/auth/acl.js
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * @fileoverview Access Control Lists
+ */
+const config = require('config');
+const NodeAcl = require('acl');
+
+const userRoles = config.roles.config;
+
+// memoryBackend starts with a lowercase from node-acl
+const acl = new NodeAcl(new NodeAcl.memoryBackend()); // eslint-disable-line new-cap
```
Snugug: So, found your problem with everyone being unloaded on server crash. Using memory backend. Should use [knex backend](https://github.com/christophertrudel/node_acl_knex)

lib/auth/acl.js
Snugug: Where is `cfg` being passed in?

lib/init/parser.js
Snugug: I understand the bug. `connect-multiparty` does the same thing, and will give `req.body.email` as well.

lib/routes/users.js
```diff
@@ -0,0 +1,225 @@
...
+    const app = application;
+
+    /*
+     * @name Users landing
+     *
+     * @param {object} req - HTTP Request
+     * @param {object} res - HTTP Response
+     * @param {object} next - Express callback
+    */
+    app.get('/users', acl.middleware(1, getUserId, 'view'), (req, res, next) => {
```
Snugug: Not pulling from config

lib/routes/users.js
```diff
@@ -0,0 +1,225 @@
...
+    });
+
+    /*
+     * @name Delete user POST method
+     *
+     * @param {object} req - HTTP Request
+     * @param {object} res - HTTP Response
+     * @param {object} next - Express callback
+    */
+    app.post('/users/delete', [multipartMiddleware, acl.middleware(1, getUserId, 'delete')], (req, res, next) => {
```
Snugug: This inline array makes me sad. Can we split it up?

lib/users/users.yml
```diff
@@ -10,7 +10,14 @@ attributes:
...
     description: Make it secure!
-  - type: text
-    id: level
-    name: Access Level
-    description: (make multi-select plugin w/selection of CTs for `content-creator`) CHOOSE one - superuser/admin/content-creator
+  - type: select
+    id: role
+    name: Authorization role
+    label: Select authorization level
+    inputs:
```
Snugug: These need to pull from the config. We're probably going to need to make this a `.js` file, not a `.yml` file to accomidate that

package.json
```diff
@@ -19,7 +19,7 @@
     "pretest": "npm run lint",
     "test": "NODE_APP_INSTANCE=test npm run nyc",
     "posttest": "gulp build",
-    "start": "nodemon -e html,js ./index.js",
+    "start": "DEBUG=express:* nodemon -e html,js ./index.js",
```
Snugug: We don't want DEBUG in there; this is what's run in production

tests/server.js
```diff
@@ -1,63 +1,87 @@
...
+  },
+];
+
+const userCheck = (usr) => {
+  database.select('*').from('users').where({
+    email: usr.email,
+  })
+  .then(user => {
+    if (user.length < 1) {
+      database('users').insert({
```
Snugug: Errors likely eaten because this is not returned

tests/server.js
```diff
@@ -1,63 +1,87 @@
...
+    password: 'pa55word',
+    role: 'creator',
+  },
+];
+
+const userCheck = (usr) => {
+  database.select('*').from('users').where({
+    email: usr.email,
+  })
+  .then(user => {
```
Snugug: Up a line please

tests/server.js
```diff
@@ -1,63 +1,87 @@
...
+    if (user.length < 1) {
+      database('users').insert({
+        email: usr.email,
+        password: bcrypt.hashSync(usr.password),
+        role: usr.role,
+      })
+      .then(() => {
```
Snugug: Attach to `database.select('*')`

tests/server.js
```diff
@@ -1,63 +1,87 @@
...
-      t.is(err, null, 'Should not have an error');
-      t.is(res.status, 200, 'should return status 200');
-      t.regex(res.text, /DOCTYPE html/, 'should have an html doctype');
-      t.end();
+// cue up our application for all tests
+test.cb.beforeEach((t) => {
+  t.context.agent = request.agent(); // eslint-disable-line
+  punchcard().then(app => {
+    accounts.forEach((usr) => {
+      userCheck(usr);
```
Snugug: `userCheck` is a promise. Needs Bluebirds's `Promise.map` method, not simple `forEach`

tests/server.js
```diff
@@ -1,63 +1,87 @@
...
-test.cb('Content Type Landing Page', t => {
-  t.context.request
-    .get('/content/services')
-    .end((err, res) => {
-      t.is(err, null, 'Should not have an error');
-      t.is(res.status, 200, 'should return status 200');
-      t.regex(res.text, /DOCTYPE html/, 'should have an html doctype');
-      t.end();
-    });
+    request(app)
```
Snugug: Because `userCheck` is a promise, going to have a race condition with this line.

tests/server.js
```diff
@@ -1,63 +1,87 @@
...
+      .post('/login')
+      .send({
+        username: accounts[0].email,
+        password: accounts[0].password,
+      })
+      .end((err, res) => { // eslint-disable-line
+        if (err) return err;
+        t.context.agent.saveCookies(res); // eslint-disable-line
```
Snugug: Using `_.set` resolves these lint issues

tests/server.js
```diff
@@ -1,63 +1,87 @@
...
-      t.regex(res.text, /DOCTYPE html/, 'should have an html doctype');
-      t.end();
-    });
+    request(app)
+      .post('/login')
+      .send({
+        username: accounts[0].email,
+        password: accounts[0].password,
+      })
+      .end((err, res) => { // eslint-disable-line
```
Snugug: Why is this line disabled?

tests/users.js
```diff
@@ -1,5 +1,6 @@
 import test from 'ava';
 import users from '../lib/users';
+import userRoutes from '../lib/routes/users';
```
Snugug: `routes`

tests/users.js
```diff
@@ -17,8 +18,8 @@ const model = {
 };
 
 test('Exports exist', t => {
-  t.is(typeof users.routes, 'function', 'Submodule `routes` exists and is a function');
-  t.is(typeof users, 'function', 'Submodule `model` exists, is the primary for `users`, and is a function');
+  t.is(typeof userRoutes, 'function', 'Submodule `routes` exists and is a function');
```
Snugug: `routes`

views/_nav.html
```diff
@@ -3,5 +3,6 @@
     {% for item in nav %}
     <li><a href="/{{item.path}}">{{item.label}}</a></li>
     {% endfor %}
+    <li><a href="/logout">logout</a></li>
```
Snugug: Only should be there if logged in

Snugug: (can add roles and auth state to locals)

tests/server.js
```diff
@@ -1,63 +1,87 @@
...
-  t.context.request
-    .get('/')
-    .end((err, res) => {
-      t.is(err, null, 'Should not have an error');
-      t.is(res.status, 200, 'should return status 200');
-      t.regex(res.text, /DOCTYPE html/, 'should have an html doctype');
-      t.end();
+// cue up our application for all tests
+test.cb.beforeEach((t) => {
+  t.context.agent = request.agent(); // eslint-disable-line
```
Snugug: Appears from [supertest](https://github.com/visionmedia/supertest) that if you're going to use `agent`, needs to be initialized as `request.agent(app);` and have that be used instead of straight `request`

Snugug: Also appears as if `supertest` has an assertion library of its own that maybe we look at, but will see if this does what we expect

PR #48, opened 2016-06-02
Users

Snugug:
User Authentication and Roles.

**Changelog**

```markdown
**New**

* User Authentication
* Roles
```

---
Resolves: #44 
Resolves #37 


Snugug:
Don't merge yet

Snugug:
This resolves all of my comments from #44 except `acl-knex` which seems to be unmaintained and currently doesn't work with ACL 4.7+. Worth investigating, later

Snugug:
This is good to review. I'd like to be able to test both admin and content creator access on these URLs, but we can look at that later

PR #60, opened 2016-06-08
Feature/content edit

scottnath:
@Snugug @rachelnicole 

Content editing.

**Changelog**

```markdown
**New**

* Edit content
* primary attribute on edit manage page
```

---

### DIFF
--- a/lib/routes/content.js
+++ b/lib/routes/content.js
@@ -180,7 +180,6 @@ const routes = application => {
 
       // something went wrong on save:
       if (Object.keys(values).length > 0) {
-
         // add the previous session data back in
         _.set(req.session, 'form.content.add', {
           [req.params.type.toLowerCase()]: {
@@ -188,7 +187,8 @@ const routes = application => {
             revision: revisionSess,
           },
         });
-        content.only(req.params.type.toLowerCase(), values, [type]).then(merged => {
+
+        return content.only(req.params.type.toLowerCase(), values, [type]).then(merged => {
           return content.form(merged, errors).then(form => {
             res.render('content/add', {
               form,
@@ -201,8 +201,9 @@ const routes = application => {
           next(e);
         });
       }
-      else {
 
+      // eslint mad if no return, then mad at this else if it is there
+      else { // eslint-disable-line no-else-return
         // Search for the revision
         return revisionSearch.then(rows => {
           if (rows.length < 1) {
@@ -216,8 +217,8 @@ const routes = application => {
           const revision = rows[0].revision;
 
           return database(`content-type--${type.id}`).where({
-            'id': req.params.id,
-            'revision': revision,
+            id: req.params.id,
+            revision,
           });
         }).then(rows => {
           if (rows.length < 1) {
--- a/lib/routes/content.js
+++ b/lib/routes/content.js
@@ -65,7 +65,7 @@ const routes = application => {
       // TODO: create node in content-type conf
       type.primary = type.attributes[0].inputs[Object.keys(type.attributes[0].inputs)[0]].name;
 
-      database
+      return database
         .distinct(database.raw('ON (id) id'))
         .select('*')
         .from(`content-type--${type.id}`)
@@ -78,8 +78,6 @@ const routes = application => {
             config: config.content,
           });
         });
-
-      return true;
     });
 
     /*
@@ -107,7 +105,7 @@ const routes = application => {
         return next();
       }
 
-      content.only(req.params.type.toLowerCase(), values, [type]).then(merged => {
+      return content.only(req.params.type.toLowerCase(), values, [type]).then(merged => {
         return content.form(merged, errors).then(form => {
           res.render('content/add', {
             form,
@@ -119,8 +117,6 @@ const routes = application => {
       }).catch(e => {
         next(e);
       });
-
-      return true;
     });
 
     /*
@@ -324,7 +320,7 @@ const routes = application => {
           approval = req.body.approval;
         }
 
-        database(`content-type--${req.params.type.toLowerCase()}`).insert({
+        return database(`content-type--${req.params.type.toLowerCase()}`).insert({
           id,
           language,
           sunrise,
@@ -345,7 +341,9 @@ const routes = application => {
           next(e);
         });
       }
-      else {
+
+      // eslint mad if no return, then mad at this else if it is there
+      else { // eslint-disable-line no-else-return
         const referrer = req.get('Referrer');
         _.set(req.session, 'form.content.add', {
           [req.params.type.toLowerCase()]: {
--- a/lib/routes/content.js
+++ b/lib/routes/content.js
@@ -4,7 +4,6 @@
  * @fileoverview Content system routing
  *
  */
-
 const config = require('config');
 const content = require('punchcard-content-types');
 const multipart = require('connect-multiparty');
@@ -12,7 +11,7 @@ const uuid = require('uuid');
 const _ = require('lodash');
 
 const utils = require('../utils');
-const db = require('../database');
+const database = require('../database');
 
 const multipartMiddleware = multipart();
 
@@ -28,8 +27,11 @@ const routes = application => {
     const types = app.get('content-types');
 
     /*
-     * Content Home Page
-     */
+     * @name Content Home Page
+     *
+     * @param {object} req - HTTP Request
+     * @param {object} res - HTTP Response
+    */
     app.get(`/${config.content.base}`, (req, res) => {
       res.render('content/home', {
         content: {
@@ -41,8 +43,12 @@ const routes = application => {
     });
 
     /*
-     * Individual Content Type Landing Page
-     */
+     * @name Individual Content Type Landing Page
+     *
+     * @param {object} req - HTTP Request
+     * @param {object} res - HTTP Response
+     * @param {object} next - Express callback
+    */
     app.get(`/${config.content.base}/:type`, (req, res, next) => {
       const type = utils.singleItem('id', req.params.type.toLowerCase(), types);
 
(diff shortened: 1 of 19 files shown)


Snugug:
Coverage failed. Please add some more tests

lib/routes/content.js
```diff
@@ -41,34 +43,46 @@ const routes = application => {
...
 
         return next();
       }
 
-      res.render('content/landing', {
-        content: {
-          base: req.url,
-          actions: config.content.actions,
+      // TEMP: use the first input of the first attribute as the primary title
+      type.primary = type.attributes[0].inputs[Object.keys(type.attributes[0].inputs)[0]].name;
```
Snugug: Where is this used?

Snugug: nm, below

lib/routes/content.js
```diff
@@ -103,14 +118,92 @@ const routes = application => {
...
-     *  Post to Content Type
-     */
-    app.post(`/${config.content.base}/:type/${config.content.actions.save}`, multipartMiddleware, (req, res, next) => {
+     * @name Individual Content Type Edit Page
+     *
+     * @param {object} req - HTTP Request
+     * @param {object} res - HTTP Response
+     * @param {object} next - Express callback
+    */
+    app.get(`/${config.content.base}/:type/:id/${config.content.actions.edit}`, (req, res, next) => {
```
Snugug: What is the page for `base/:type/:id`? Is that a read-only version of the form? Does it redirect to `edit`?

Snugug: Maybe lots of metadata about that particular id? Like sunrise/sunset, publishable, workflow state, etc…?

lib/routes/content.js
```diff
@@ -103,14 +118,92 @@ const routes = application => {
...
+      const type = utils.singleItem('id', req.params.type.toLowerCase(), types);
+
+      // need to merge errors for edit/revisions
+      // const errors = _.get(req.session, 'form.content.edit.errors', {});
```
Snugug: Need to do this

lib/routes/content.js
```diff
@@ -41,34 +43,46 @@ const routes = application => {
...
       }
 
-      res.render('content/landing', {
-        content: {
-          base: req.url,
-          actions: config.content.actions,
+      // TEMP: use the first input of the first attribute as the primary title
+      type.primary = type.attributes[0].inputs[Object.keys(type.attributes[0].inputs)[0]].name;
+
+      database.select('*').from(`content-type--${type.id}`).then(rows => {
```
Snugug: Why can't you return your promise now that you have something to return?

lib/routes/content.js
```diff
@@ -41,34 +43,47 @@ const routes = application => {
...
 
-      res.render('content/landing', {
-        content: {
-          base: req.url,
-          actions: config.content.actions,
+      // TEMP: use the first input of the first attribute as the primary title
+      // TODO: create node in content-type conf
+      type.primary = type.attributes[0].inputs[Object.keys(type.attributes[0].inputs)[0]].name;
+
+      database.select('*').from(`content-type--${type.id}`).then(rows => {
```
Snugug: Why can't you return this promise now that you have it, instead of the `true`

lib/routes/content.js
```diff
@@ -103,14 +119,92 @@ const routes = application => {
...
+
+          return next();
+        }
+
+        const value = rows[0].value;
+
+        // mapping data from database
+        Object.keys(value).map(key => {
+          // if sunrise/sunset object is structured differently
+          if ((key === 'sunset-date') || (key === 'sunset-time') || (key === 'sunrise-date') || (key === 'sunrise-time')) {
```
Snugug: I don't understand this. It should _always_ be the same

lib/routes/content.js
```diff
@@ -103,14 +119,92 @@ const routes = application => {
...
+            data[key] = {
+              'value': value[key],
+            };
+          }
+          else {
+            const input = key.split('--');
```
Snugug: content-types [includes the split functionality](https://github.com/punchcard-cms/content-types/blob/master/lib/form/validate.js#L22-L50) in `validate` and supports repeatables. Maybe we should expose that instead of re-writing that functionality?

lib/routes/content.js
```diff
@@ -152,7 +246,17 @@ const routes = application => {
       }
 
       return true;
-    });
+    };
+
+    /*
+     * @name Add - Post to Content Type
+    */
+    app.post(`/${config.content.base}/:type/${config.content.actions.save}`, multipartMiddleware, save);
```
Snugug: I literally don't understand why you are making multiple endpoints that do the exact same thing. They should all be calling `content/save`, that's why I made it like that

tests/server.js
```diff
@@ -84,6 +117,14 @@ test.cb.before((t) => {
   });
 });
 
+test.cb.after((t) => {
+  punchcard().then(() => {
```
Snugug: Gotta return your promises

tests/server.js
```diff
@@ -42,11 +44,42 @@ const userCheck = (usr) => {
...
+const deleteService = () => {
+  return database('content-type--services').where('id', '=', serviceUuid).del();
+};
+
 // cue up our application for all tests
 test.cb.before((t) => {
   punchcard().then(app => {
     agent = request.agent(app);
 
+    return Promise.map(service, serviceCheck);
```
Snugug: If this only happens in `before`, just put it all in `before`

lib/routes/content.js
scottnath: @Snugug for this one, the comment is missing a comment. Should be:

```
// if sunrise/sunset, object is structured differently
```

tests/server.js
scottnath: @Snugug, @Snugug said it is ok to not move this function (serviceCheck) into `before` and to disregard the original comment by @Snugug

tests/server.js
scottnath: returning these promises breaks the tests

lib/routes/content.js
scottnath: will do

scottnath:
@scottnath: note-to-self: first, fix bug with `./tests/server.js` not deleting test entries in the service content's database table.

tests/server.js
Snugug: Then we won't do it! Live, from Hackademy!

lib/routes/content.js
```diff
@@ -144,60 +153,136 @@ const routes = application => {
...
+        });
 
-          return next();
-        }
+        return next();
+      }
+
+      // no revision in url, so create search to find latest revision
+      if (!req.params.revision || !Number.isInteger(req.params.revision)) {
+        revisionSearch = database(`content-type--${type.id}`)
```
Snugug: These should be promises, no? Can't just set the result to the output

lib/routes/content.js
```diff
@@ -144,60 +153,136 @@ const routes = application => {
...
+          .select('revision')
+          .where('id', req.params.id)
+          .orderBy('revision', 'DESC')
+          .limit(1);
+      }
+      else {
+        revisionSearch = database(`content-type--${type.id}`)
```
Snugug: Same as above

lib/routes/content.js
```diff
@@ -65,14 +65,21 @@ const routes = application => {
       // TODO: create node in content-type conf
       type.primary = type.attributes[0].inputs[Object.keys(type.attributes[0].inputs)[0]].name;
 
-      return database.select('*').from(`content-type--${type.id}`).then(rows => {
-        res.render('content/landing', {
-          title: config.content.base,
-          content: rows,
-          type,
-          config: config.content,
+      database
```
Snugug: This return disapeared

lib/routes/content.js
```diff
@@ -144,60 +153,136 @@ const routes = application => {
...
+      if (Object.keys(values).length > 0) {
+
+        // add the previous session data back in
+        _.set(req.session, 'form.content.add', {
+          [req.params.type.toLowerCase()]: {
+            id: idSess,
+            revision: revisionSess,
+          },
+        });
+        content.only(req.params.type.toLowerCase(), values, [type]).then(merged => {
```
Snugug: Return?

lib/routes/content.js
```diff
@@ -215,20 +300,45 @@ const routes = application => {
...
+          language = req.body.language;
+        }
+
+        // approval
+        let approval = 0;
+        if (req.body.hasOwnProperty('approval')) {
+          approval = req.body.approval;
+        }
+
         database(`content-type--${req.params.type.toLowerCase()}`).insert({
```
Snugug: Gotta return your promises

Snugug:
@scottnath Comments in. Please update and then we can go

lib/routes/content.js
```diff
@@ -41,58 +43,77 @@ const routes = application => {
...
-          message: config.content.messages.missing.replace('%type', req.params.type),
+          message: config.content.messages.missing.type.replace('%type', req.params.type),
           safe: `/${config.content.base}`,
         });
 
         return next();
       }
 
-      content.only(req.params.type, values, [type]).then(merged => {
+      content.only(req.params.type.toLowerCase(), values, [type]).then(merged => {
```
Snugug: Return yo promises

lib/routes/content.js
```diff
@@ -41,58 +43,77 @@ const routes = application => {
...
-          base: req.url,
-          actions: config.content.actions,
-          type,
-        },
-      });
+      // TEMP: use the first input of the first attribute as the primary title
+      // TODO: create node in content-type conf
+      type.primary = type.attributes[0].inputs[Object.keys(type.attributes[0].inputs)[0]].name;
+
+      database
```
Snugug: Return yo promises

lib/routes/content.js
```diff
@@ -123,20 +301,45 @@ const routes = application => {
...
+          language = req.body.language;
+        }
+
+        // approval
+        let approval = 0;
+        if (req.body.hasOwnProperty('approval')) {
+          approval = req.body.approval;
+        }
+
+        database(`content-type--${req.params.type.toLowerCase()}`).insert({
```
Snugug: Return yo promises

lib/routes/content.js
scottnath: @Snugug these are promises. [See below for usage as promises](https://github.com/scottnath/punchcard/blob/e4112ffadf3ea1274c8a3161dbabf63bd1a3e949/lib/routes/content.js#L208)
I'm giving `revisionSearch` either a promise that searches for a revision if one isn't provided, or a promise to test that the revision they provided actually exists (line 175 below)

scottnath:
@Snugug responses in. Please re-review and then we can go

Go where? Who knows?

lib/routes/content.js
```diff
@@ -345,7 +341,9 @@ const routes = application => {
           next(e);
         });
       }
-      else {
+
+      // eslint mad if no return, then mad at this else if it is there
+      else { // eslint-disable-line no-else-return
```
Snugug: Grumpy ESLint

PR #105, opened 2016-07-06
Feature/workflows

### LINKED ISSUE #98: Core Developer: Content Form
Workflow UX for the content form

```gherkin
Feature: Content form
  As a core developer
  I want to make changes to the content edit/add form
  So that the form has workflow functionality

  Scenario: Workflow button
    Given that there needs to be a CTA for workflow
    When I update the content form
    Then there should be a button for the first step in the workflow

  Scenario: Basic workflow
    Given that the basic workflow allows a Content Creator to self-publish
    When only the basic workflow is required by a content type
    Then the workflow button should say 'Publish'
      And clicking that button should make the content publish-able.

  Scenario: Non-basic workflow
    Given that non-basic workflow requires the user to submit content for approval
    When non basic workflow is present
    Then the workflow button should say 'submit for approval'
      Or the text of this button could come from the workflow config.
```

---

- [x] content form / button
- [x] content form / basic
- [x] content form / non-basic

scottnath:
Workflows flow is in.

---
Resolves #98 

gladiator


lib/database/schemas.js
```diff
@@ -58,6 +58,10 @@ const content = [
       update: 'CASCADE',
     },
   },
+  {
+    name: 'workflows',
+    type: 'jsonb',
+  },
```
Snugug: What is this for? Is this for the comments/audit trail?

scottnath: yes. comments/audit trail.

Object created for that entry: https://github.com/scottnath/punchcard/blob/f247101e6d96eb0553804c8c56d5930c939f3046/lib/routes/workflows.js#L163

`workflows` will have an array of those objects.

Snugug: Don't like the name `workflows`, can it be something like `approval trail` or `approval` or `audit` or something?

lib/workflows/load.js
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * @returns {Promise} - An {Array} containing {Object} for each workflow, including `name` {String}, `id` name {String}, and steps {Object} loaded from YAML config file
+ */
+const load = () => {
+  const workflows = [];
+  const ids = [];
+  let workflowsPath = path.join(process.cwd(), 'workflows');
+
+  if (globalConfig.hasOwnProperty('workflows')) {
+    if (globalConfig.content.hasOwnProperty('workflows')) {
```
Snugug: This looks like you copied it directly out of Content Types. That doesn't look like what you want, based on the actual config you set up

config/default.js
```diff
@@ -21,10 +21,22 @@ module.exports = {
     },
     directory: path.join(__dirname, '../content-types'),
     messages: {
+      format: {
+        revision: 'Revision must be a number',
```
Snugug: This crap is ugly and I don't like it and we really _really_ need to get [i18n](https://www.npmjs.com/package/i18n) in to do away with all of this

lib/workflows.js
```diff
@@ -0,0 +1,21 @@
...
+const model = require('./workflows/data');
+
+module.exports = model;
+
+/**
+ * Get all Workflows
+ *
+ * @returns {Promise} - An {Array} containing {Object} for each workflow, including `name` {String}, `id` name {String}, and steps {Object} loaded from YAML config file
+ */
+const raw = loader;
```
Snugug: Just set this directly

lib/workflows.js
```diff
@@ -0,0 +1,21 @@
+/**
+ *  @fileoverview User functions index
+ *
+ *  @author  Scott Nath
+ *
+ */
+'use strict';
+
+const loader = require('./workflows/load');
+const model = require('./workflows/data');
```
Snugug: Why are you changing names around here, when it's so simple to have them be the same

lib/workflows/load.js
Snugug: It should be fairly trivial to make workflows pluggable at this point, why not just do it? Then we can distribute self-publish and editor-approve immediately

Snugug:
https://github.com/scottnath/punchcard/blob/f247101e6d96eb0553804c8c56d5930c939f3046/lib/routes/content.js#L377-L381
On save need to set approval to the length of the workflow, not 0

lib/routes/content.js
```diff
@@ -126,14 +164,15 @@ const routes = application => {
      * @param {object} res - HTTP Response
      * @param {object} next - Express callback
     */
-    app.get(`/${config.content.base}/:type/:id/:revision/${config.content.actions.edit}`, (req, res, next) => {
+    app.get(`/${config.content.base}/:type/:revision/${config.content.actions.edit}`, (req, res, next) => {
```
Snugug: Needs to be `/:id/:revision` (I know we only need to look up by revision but we need the ID for user s to be able to get to the listing of all IDs. REST architecture. Applies to workflows too)

lib/routes/workflows.js
```diff
@@ -0,0 +1,206 @@
...
+ * @fileoverview Content system routing
+ *
+ */
+const config = require('config');
+const content = require('punchcard-content-types');
+const _ = require('lodash');
+
+const utils = require('../utils');
+const database = require('../database');
+const flow = require('../workflows/data');
```
Snugug: You messing with these names makes the code very hard to follow

Snugug: This should be `workflows` and it should be requiring `../workflows`; that'll give you `data`

lib/routes/workflows.js
```diff
@@ -0,0 +1,206 @@
...
+              name: input[0],
+              value: revision.value[key],
+            };
+            values.push(data);
+          });
+
+          return flow().then(workflow => {
+            return content.only('workflow', {}, workflow).then(wf => {
+              return content.form(wf).then(form => {
+                res.render('content/approval', {
```
Snugug: It appears as if there is a single approver URL for all steps in a workflow? There needs to be a separate URL per step in the workflow. Two steps? `content/type/id/revision/step-1`, `content/type/id/revision/step-2` (or whatever the final structure is going to be)

Snugug: May need the `approve` or `workflow` keyword in there somewhere, but it would need to make sense at a top-level landing page. Maybe something like `{{content}}/{{type}}/approval/{{id}}/{{revision}}/step` so `{{content}}/{{type}}/approval` is all of the revisions available for approval, plus `id` then `revision`.

Now that I'm saying this out loud, may not need the actual step name in the URL, just want to make sure the step isn't being chosen from a select or something

Snugug:
What you have is what you're going to get from me for right now.

scottnath:
@Snugug waiting on tests...enjoy

Snugug:
@scottnath for things you're testing, `:crystal_ball:` 🔮

scottnath:
aha...didn't know `:crystal_ball:`. will do now

lib/workflows/load.js
```diff
@@ -0,0 +1,66 @@
+'use strict';
+
+const dir = require('node-dir');
+const yaml = require('js-yaml');
+const kebab = require('lodash/kebabCase');
```
Snugug: As we discussed in API, we should use `underscore.string` instead of lodash's `kebabCase`

lib/workflows/load.js
```diff
@@ -0,0 +1,66 @@
...
+    }, (err) => {
+      if (err) {
+        reject(err);
+      }
+      resolve(workflows);
+    });
+  });
+};
+
+module.exports = load;
```
Snugug: This still hasn't been resolved it seems

lib/workflows/workflow.js
```diff
@@ -0,0 +1,43 @@
...
+    {
+      type: 'select',
+      id: 'action',
+      name: 'Action',
+      required: 'save',
+      description: 'Please choose to approve or reject this content.',
+      inputs: {
+        select: {
+          options: [
+            { label: 'Select Action',
```
Snugug: Why is `Select Action` an option? Why not a label?

package.json
```diff
@@ -81,6 +86,7 @@
     "input-plugin-email": "0.1.0",
     "input-plugin-text": "0.0.5",
     "jsdom": "^9.2.0",
+    "lodash": "^4.13.1",
```
Snugug: Lodash is already a dependency on line 65

src/sass/partials/global/_extends.scss
```diff
@@ -0,0 +1,26 @@
+.base--a {
```
Snugug: We are not using extends. These have styling elsewhere (in base) that you should use

Snugug:
@scottnath as a reminder, you need to include the issue(s) this PR closes at the top, and sign off, before I can merge this

scottnath:
@Snugug finally complete! Increased code coverage!

lib/workflows.js
```diff
@@ -5,15 +5,99 @@
...
+ * @param {object} req - HTTP Request
+ *
+ * @returns {AuditEntry} audit entry
+ */
+const entry = (comment, action, step, req) => {
+  const data = {
+    comment,
+    action,
+    step,
+    author: req.user.id,
```
Snugug: Can we change this from `user.id` to either be an object that contains the `id` and the `email`, or just the `email` of the user? If a user is deleted from the system, the `id` is meaningless for the audit trail

lib/workflows.js
```diff
@@ -5,15 +5,99 @@
...
+ * @params {object} revision
  */
-module.exports.raw = load;
+const audits = (revision, workflow, req) => {
+  const rev = revision;
+  const comment = req.body['comment--textarea'];
+  let audits = rev.audit;
+  let step;
+  let approval;
+  let publishable = false;
```
Snugug: Not a blocker, but I like to see all variables that have values assigned defined before those that are just initialted

lib/utils.js
```diff
@@ -209,10 +210,35 @@ const format = body => {
   return data;
 };
 
+/*
+ * Get a single workflow from all workflows available
+ */
+const workflow = (type, allflows, globalConfig, req) => {
```
Snugug: This should live in the `workflows` file not the `utils` file

lib/routes/content.js
```diff
@@ -437,3 +484,4 @@ const routes = application => {
 };
 
 module.exports = routes;
+module.exports.check = check;
```
Snugug: If you find yourself needing to export an additional function from these route files, it likely means it doesn't belong in this file and should be moved to a separate, related file

Snugug: So please do that

lib/routes/workflows.js
```diff
@@ -0,0 +1,196 @@
...
+                .update(audits);
+            }).then(() => {
+              _.unset(req.session, 'form.content.approve');
+              res.redirect(`/${config.content.base}/${req.params.type}`);
+            }).catch(e => {
+              next(e);
+            });
+          }
+
+          // eslint thinks the return above ruins this else
```
Snugug: It does. That rule says that if there is a return inside of an `if`, you shouldn't wrap the remaining `else` in an `if` statement

Snugug: If it breaks when you remove it, that's one story, but this rule shoud be correct

lib/workflows/load.js
Snugug: This plz

lib/workflows/load.js
Snugug: This plz

package.json
```diff
@@ -19,6 +19,7 @@
     "test": "NODE_APP_INSTANCE=test npm run nyc",
     "posttest": "gulp build",
     "start": "nodemon -e html,js ./index.js",
+    "debug": "nodemon -e html,js --debug ./index.js",
```
Snugug: Debug should be added to the Gulp nodemon runner, no?

package.json
Snugug: This plz

src/sass/partials/global/_extends.scss
Snugug: This plz

Snugug:
Got some new comments, and some that weren't dealt with from before.

I also see you exporting a lot of functions all the way through the pipeline (`workflow.structure` for instance) that are not needed by anything in the actual system except testing. Exporting from their single file _if_ it cannot be touched in testing otherwise is OK, but no need to export them all the way through

src/sass/partials/global/_extends.scss
scottnath: whoops. slipped through again.

lib/workflows/workflow.js
scottnath: This is a UX issue actually. 

problem:
I need to have the form return either `null`, `approve`, or `reject`. If I don't have an option for `value=""` then there's no-way to have a `select`. I could use radios instead - with no radio selected; or some other idea that might need specific validation because `required` validation only checks for an empty string.

solution:
I was hoping to get this in and have a discussion as to the best way to get this selection (reject or approve) into the form and have it work, without javascript, while at the same time using content-types native validation function.

Could be radios, or a select. Or maybe there are some other idea (should be discussed though)

lib/routes/workflows.js
scottnath: fixed. sometimes I'm slow

lib/workflows/load.js
scottnath: error checking now in `utils.check()`

scottnath:
@Snugug all requested changes added. Waiting on tests, which worked locally.

lib/content/utils.js
```diff
@@ -0,0 +1,76 @@
+'use strict';
+
+const _ = require('lodash');
+const isUUID = require('validator/lib/isUUID');
+
+
+const check = {
+
+  /**
```
Snugug: Not blocking, but I'd prefer these to be top-level functions that get bundled together in to `check` in the `module.exports`

lib/content/utils.js
```diff
@@ -0,0 +1,76 @@
...
+   * Determine if an id is in uuid format; update session for error
+   *
+   * @param  {object} config global config
+   * @param {object} req - HTTP Request
+   *
+   * @returns {boolean} false-not in uuid format / true-it is!
+   */
+  id: (config, req) => {
+    if ((typeof req.params.id !== 'string') || (!isUUID(req.params.id))) {
+      _.set(req.session, '404', {
```
Snugug: This is the very definition of a side effect. We don't want magic side effects. Can we please have this just return true/false (or true/message) and have a separate helper that will set the 404 redirect? Especially as the 404 redirect is _identical_ except for the message?

lib/content/utils.js
```diff
@@ -0,0 +1,76 @@
...
+
+  /**
+   * Determine if an id is in uuid format; update session for error
+   *
+   * @param  {object} config global config
+   * @param {object} req - HTTP Request
+   *
+   * @returns {boolean} false-not in uuid format / true-it is!
+   */
+  id: (config, req) => {
```
Snugug: Instead of passing `config` in to each of these, we should _use the global config_ that everything else is using

lib/workflows/utils.js
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  if (!config.hasOwnProperty('steps')) {
+    return 'A workflow must have steps';
+  }
+
+  if (!Array.isArray(config.steps)) {
+    return 'Workflow steps must be an array';
+  }
+
```
Snugug: Still doesn't check that each step is an object with a `name` property and that, if it has a `self` property, that that is a boolean

Snugug:
@scottnath Still some comments. Once those are resolved and tests pass we should be just about there

lib/content/utils.js
scottnath: @Snugug sentence phrasing confusing. So, for `bundled together in to check in the module.exports`, I think you mean that there will be separate functions, but they would be called like this:

```
module.exports = {
  check: {
    id,
    revision,
    type,
  },
  other,
  functions,
}
```

correct?

lib/content/utils.js
scottnath: While I agree, this will cause a structural change across other files not related to the workflows PR. This works now, can we leave it and make a note? I can clean this up when I clean up users (which needs to happen to get role-level functions for workflows)

Snugug: Please create an issue and report back here

Snugug:
The API tests failed, probably a `deepEqual` thing. You may need to fix to strings just to get this to reliably pass, sorry :\
Running again to see if we can get a clean

lib/content/utils.js
scottnath: https://github.com/punchcard-cms/punchcard/issues/142

lib/content/utils.js
scottnath: removed

lib/workflows/utils.js
scottnath: does now. it's amazing.

scottnath:
@Snugug I may have skipped the offending API tests in this PR - please check.

Regardless, all changes are hopefully complete with the latest push.

lib/workflows/utils.js
```diff
@@ -34,16 +34,35 @@ const check = (config) => {
...
+    }
     if (step.self && typeof step.self !== 'boolean') {
-      return 'Whether a step is self-publish must be a boolean';
+      return 'Self-publish must be a boolean';
+    }
+
+    return true;
+  }).filter((res) => {
+    if (res !== true) {
+      return res;
```
Snugug: Filter should return `true` or `false`, not a truthy thing like a string. Filter does not manipulate its array's value, only determines if it should be in the array.

Snugug:
ALL. LOOKS. GOOD.

Snugug:
Except for that one tiny comment about filters. If you could fix that, this will be merged. Thanks!

lib/workflows/utils.js
scottnath: Totally understand how filter works. I am determining what I do _not_ want in this array - ie, all the `true` ones. I am using `map` to test all of the steps, then I use `filter` to...wait for it...*_filter_* out the `true` ones, leaving me with either an array of strings (of which I return the first) or an empty array, wherein I return the truest true that was ever true.

The return of the string, although functional, is confusing though. I'll switch it to true.

Here is a codepen breaking it down: http://codepen.io/scottnath/pen/EyRaWy?editors=1111

Snugug:
Who's a happy @scottnath?!

poofichu:
😂😂😂

PR #110, opened 2016-07-11
New structure to store form data into DB

ayush2k:
`DCO 1.1 Signed-off-by: {{Ayush Gupta} <{{ayush2k@gmail.com}}>`

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -127,10 +127,48 @@ const config = values => {
   return result;
 };
 
+// Structures input from form
+const format = body => {
+  const data = {};
+  Object.keys(body).forEach(key => {
+    if ((key === 'sunset-date') || (key === 'sunset-time') || (key === 'sunrise-date') || (key === 'sunrise-time')) {
+      data[key] = {
+        'value': body[key],
+      };
+    }
+    else {
+      const inputs = key.split('--');
+      const inputName = inputs[0];
+      const inputType = inputs[1];
+
+      // Checks if multiple instances
+      if (inputs.length >= 3) {
+        const index = inputs[2];
+        if (!data.hasOwnProperty(inputName)) {
+          data[inputName] = [];
+        }
+        if (!data[inputName][index]) {
+          // Creates an array for multiple instances
+          data[inputName][index] = {};
+        }
+        data[inputName][index][inputType] = { 'value': body[key] };
+      }
+      else {
+        if (!data.hasOwnProperty(inputName)) {
+          data[inputName] = {};
+        }
+        data[inputName][inputType] = { 'value': body[key] };
+      }
+    }
+  });
+
+  return data;
+};
 
 module.exports = {
   singleItem,
   log,
   time,
   config,
+  format,
 };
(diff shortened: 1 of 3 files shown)


ayush2k:
@Snugug @scottnath 
It fails one test

lib/utils.js
```diff
@@ -127,10 +127,48 @@ const config = values => {
   return result;
 };
 
+// Structures input from form
```
scottnath: Please add jsdoc code to the top of this. Should include a description `@param` and `@returns`

lib/utils.js
```diff
@@ -127,10 +127,48 @@ const config = values => {
...
+
+      // Checks if multiple instances
+      if (inputs.length >= 3) {
+        const index = inputs[2];
+        if (!data.hasOwnProperty(inputName)) {
+          data[inputName] = [];
+        }
+        if (!data[inputName][index]) {
+          // Creates an array for multiple instances
+          data[inputName][index] = {};
```
scottnath: You're creating an object here, not an array

tests/utils.js
```diff
@@ -122,3 +122,35 @@ test('Config', t => {
...
+
+test('Request Format', t => {
+  const input = {
+    'sunset-date': '',
+    'sunset-time': '',
+    'sunrise-date': '2016-06-23',
+    'sunrise-time': '00:00',
+    'service-name--text': 'bar',
+    'service-email--email--0': 'foo@test.com',
+    'service-email--email--1': 'bar@test.com',
```
scottnath: Please add what would be the output of a set of checkboxes or a multi-select, which should be an array.

tests/utils.js
```diff
@@ -122,3 +122,35 @@ test('Config', t => {
...
 });
+
+test('Request Format', t => {
+  const input = {
+    'sunset-date': '',
+    'sunset-time': '',
+    'sunrise-date': '2016-06-23',
+    'sunrise-time': '00:00',
+    'service-name--text': 'bar',
+    'service-email--email--0': 'foo@test.com',
```
scottnath: Do another couple of these, making some of them blank.

lib/utils.js
```diff
@@ -127,7 +127,14 @@ const config = values => {
   return result;
 };
 
-// Structures input from form
+/*
+ * Formats raw data from form to object
+ *
+ * @param {object} body - Raw data object
```
scottnath: this doesn't actually help much. Please use the [`typedef` descriptor](http://usejsdoc.org/tags-typedef.html) to break down what is in both the param and returns.

[Here is an example usage in our code](https://github.com/punchcard-cms/punchcard/blob/64921680ac6bd635725bc09908ea6a292dded73c/lib/utils.js#L63)

```
/*
 * @typedef FormattedDate
 * @type object
 *
 * @property {string} date - The date, formatted YYYY-MM-DD (e.g. 2016-05-25)
 * @property {string} time - The time, formatted HH:mm (e.g. 13:01)
 * @property {string} zone - The time zone, formatted Z (e.g. America/New_York)
 */

/*
 * Formats ISO date in to requisite components
 *
 * @param {string} date - ISO UTC formatted date
 * @param {string} zone - Timezone to retrieve date in
 *
 * @returns {FormattedDate} - The individually formatted date components
 */
```

PR #132, opened 2016-07-20
Feature/api

### LINKED ISSUE #114: REST API
Provide RESTful APIs for accessing content

**Definitions**
* Endpoint - The API URL (without hostname) to retrieve the full item.
* Slug - The sluggified version (all lowercase, underscores and spaces replaced with single dashes) of the mentioned item

```gherkin
Feature: REST API
  As an application developer
  I want to access the content in the CMS from my application
  So that I can maintain my application and its content separately

  Scenario: All Content
    Given I would like to retrieve all available content 
     When I hit the root of the API
     Then I should be given a list of all available content
      And be provided with their IDs, type, key attribute value, key attribute slug, and endpoint
      And it should be sorted by the key attribute value

  Scenario: All Content Types
    Given I would like to retrieve all of the types of content that are available
     When I hit the `types` endpoint
     Then I should be given a list of all available content types
      And be provided with their name, name slug, description, endpoint, and count of available content
      And it should be sorted by name

  Scenario: All Content of Given Content Type
    Given I would like to retrieve all of the content of a given content type that are available
     When I hit the `types/type` endpoint
     Then I should be given a list of all available content of the provided content types
      And be provided with their IDs, type, type slug, key attribute value, key attribute slug, and endpoint
      And it should be sorted by the key attribute value

  Scenario: Individual Piece of Content
    Given I would like to retrieve an individual piece of content
     When I hit the `types/type/id` endpoint
     Then I should be given the full display of requested content
      And be provided with its IDs, type, type slug, key attribute value, key attribute slug, and all attributes

Scenario: Paginate Results
    Given I am retrieving a list of items from an API
     When the list of items is larger than 30 items
     Then I should be provided results in groups of 30 items
      And I should be able to set the total number of items to return, up to 100
      And I should be able set the grouping offset I would like to retrieve
      And if I do not specify an offset I should be given the first grouping (starting at 1)
      And I should be provided with the endpoint for the next grouping
      And I should be provided with the endpoint for the previous grouping
      And I should be provided with the endpoint for the first grouping
      And I should be provided with the endpoint for the last grouping
      And it should paginate after sorting has occurred

Scenario: Sort Results
    Given I am retrieving a list of items from an API
     When the list of items is returned to me
     Then it should be sorted in ascending order by the default key of the list
      And I should be able to set the key to sort on
      And I should be able to set if sorting is ascending or descending
```

---

**REST API**
- [x] All Content
- [x] All Content Types
- [x] All Content of Given Content Type
- [x] Individual Piece of Content
- [x] Paginate Results
- [x] Sort Results

Snugug:
REST API Endpoints.

Code review can take place as I'm finishing up tests

---
Resolves #114 


Snugug:
This is OK to review. The tests seem to be a little fragile because I'm randomly generating stuff. Thoughts on that?

Snugug:
Also, most of these lines are tests, not actual functional code

lib/api/apis.js
```diff
@@ -0,0 +1,184 @@
...
+};
+
+/*
+ *  API for all Content
+ *
+ * @param {object} query - Request queries
+ *
+ * @returns {object} - All items and the pagination for the items
+ */
+// TODO: Tests
```
scottnath: nitpik: not true, right, tests exist, right? (they do, I checked); (will not keep writing this for other TODO: tests comments)

lib/api/apis.js
```diff
@@ -0,0 +1,184 @@
...
+// TODO: Return error if no items are found
+const all = query => {
+  const organize = utils.organize(query);
+
+  return database.select('*').from('live').orderBy(organize.sort.by, organize.sort.dir).offset(organize.page.offset).limit(organize.page.limit).then(rows => {
```
scottnath: we are being inconsistent with how we write this statement. Sometimes `database('table')` sometimes with `from()`. We should choose one.

lib/api/apis.js
```diff
@@ -0,0 +1,184 @@
...
+    return database('live').count('id').then(total => {
```
scottnath: Case in point

lib/api/apis.js
```diff
@@ -0,0 +1,184 @@
...
+/*
+ *  API for all Content Types
+ *
+ * @param {object} query - Request queries
+ * @param {object} cts - Content types
+ *
+ * @returns {object} - All items and the pagination for the items
+ */
+// TODO: Tests
+// TODO: Return error if no items are found
```
scottnath: agreed. Please story it up! don't see it [here](https://github.com/punchcard-cms/punchcard/issues/114)

tests/apis.js
```diff
@@ -0,0 +1,430 @@
+import test from 'ava';
+import uuid from 'uuid';
+import ipsum from 'lorem-ipsum';
+import Promise from 'bluebird';
+import slugify from 'underscore.string/slugify';
```
scottnath: why the switch from `lodash/kebabCase`? Now we have two different modules doing the same thing

tests/apis.js
```diff
@@ -0,0 +1,430 @@
...
+import _ from 'lodash/cloneDeep';
```
scottnath: we need consistency in how we call lodash as well. I would say if it's not all of lodash, then you should title the module according to the method ie: `cloneDeep`. Having `_` mean different things in different files will make maintenance complicated.

lib/database/schemas.js
```diff
@@ -1,5 +1,7 @@
 'use strict';
 
+const _ = require('lodash/cloneDeep');
```
scottnath: inconsistent usage

lib/api/utils.js
```diff
@@ -0,0 +1,172 @@
...
+};
+
+/*
+ * Determines raw live data organization
+ *
+ * @param {object} params - Request parameters
+ *
+ * @returns {object} - Object with sort key, sort direction, pagination count and offset
+ */
+const organize = (params, lookup) => {
```
scottnath: you should have some kind of check on `params` and `lookup`.

tests/apis.js
```diff
@@ -0,0 +1,430 @@
...
+      limit: 1,
+      page: 1,
+    },
+  };
+
+  t.deepEqual(actual, expected, 'Returns defaults');
+});
+
+test('Utils: Page - First', t => {
+  const organized = apiUtils.organize({});
```
scottnath: This included `{}` object could be skipped if there was error checking on `organize`

lib/api/apis.js
```diff
@@ -0,0 +1,184 @@
...
+ *  A Specific Piece of Content
+ *
+ * @param {object} query - Request queries
+ *
+ * @returns {object} - All items and the pagination for the items
+ */
+// TODO: Tests
+// TODO: Return error if no items are found
+const one = (query, id) => {
+  // const organize = utils.organize(query);
```
scottnath: detritus

lib/api/apis.js
```diff
@@ -0,0 +1,184 @@
...
+        url: `/api/types/${formatted.type_slug}`,
+      };
+
+      formatted.type = type;
+      delete formatted.type_slug;
+
+      return formatted;
+    }
+
+    return {};
```
scottnath: do you want this to be a blank object, or do you want this to be an error statement?

lib/api/apis.js
```diff
@@ -0,0 +1,184 @@
...
+
+      return formatted;
+    });
+  });
+};
+
+/*
+ *  A Specific Piece of Content
+ *
+ * @param {object} query - Request queries
```
scottnath: need `id`

lib/api/apis.js
```diff
@@ -0,0 +1,184 @@
...
+    pages,
+  };
+
+  return formatted;
+};
+
+/*
+ *  API for all Content of a given Type
+ *
+ * @param {object} query - Request queries
```
scottnath: need `type`

lib/api/utils.js
```diff
@@ -0,0 +1,172 @@
...
+const utils = require('../utils');
+
+const liveSchema = utils.singleItem('name', 'live', schemas).fields.map(field => {
+  return field.name;
+});
+
+/*
+ * Formats raw live data in to consumable returns
+ *
+ * @param {array} content - Array of objects representing the raw DB query of content
```
scottnath: need `attrs`

lib/api/utils.js
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    return result;
+  });
+};
+
+/*
+ * Determines raw live data organization
+ *
+ * @param {object} params - Request parameters
```
scottnath: need `lookup`

lib/api/utils.js
```diff
@@ -0,0 +1,172 @@
...
+  if (page < 1) {
+    page = 1;
+  }
+
+  // Ensure pagination fits within parameters
+  if (perPage < 1) {
+    perPage = 1;
+  }
+  else if (perPage > 100) {
+    perPage = 100;
```
scottnath: does this mean we have an absolute top limit of 100 per page?

lib/database/schemas.js
scottnath: see complaint above.

lib/init/authenticated.js
```diff
@@ -14,7 +14,7 @@ const config = require('config');
  *
  */
 const authenticated = (req, res, next) => {
-  if ((!req.isAuthenticated || !req.isAuthenticated()) && req.url !== config.authentication.login.path && !req.url.startsWith('/css') && !req.url.startsWith('/js') && !req.url.startsWith('/images') && !req.url.startsWith('/favicon')) {
+  if ((!req.isAuthenticated || !req.isAuthenticated()) && req.url !== config.authentication.login.path && !req.url.startsWith('/css') && !req.url.startsWith('/js') && !req.url.startsWith('/images') && !req.url.startsWith('/favicon') && !req.url.startsWith('/api')) {
```
scottnath: not for this PR, but we need to clean this up. I'll make a note!

lib/routes/api.js
```diff
@@ -0,0 +1,73 @@
...
+    /*
+     *  API for all Content Types
+     */
+    // TODO: Tests
+    // TODO: Return error if no items are found
+    app.get('/api/types', (req, res) => {
+      const formatted = api.content(req.query, types);
+
+      res.json(formatted);
+    });
```
scottnath: why no catch/next on this `get`?

scottnath:
OK. Review complete. Some fixes required.

Non-PR-fix action items:

* docs: API query examples in README
* development: jsdoc linting working?
* story-to-write: error returns
* cleanup: lib/init/authenticated.js: `authenticated` clean up caveats 
    * don't worry about this - when I do workflows' auth requirements I'll prol be in this file.

lib/api/apis.js
Snugug: We have more instances of the former than the later (and used it myself in some places), so I swapped, but FWIW I like `database.select.from` better than `database(thing).select` from a readability perspective.

TODO:
We should write an ESLint plugin to lint for this

tests/apis.js
Snugug: I searched for slugify as that's what I was expecting. I can swap to `kebabCase`

Snugug: We talked about it, going to use `underscore.string` for string manipluation instead of lodash

lib/api/apis.js
Snugug: For now, a blank object. I haven't figured out how I want to handle errors yet

lib/api/utils.js
Snugug: Yes, we max out at 100 items per page. It's the limit GitHub imposes, so I thought that was good enough for our stuff. In the Gherkin

lib/routes/api.js
Snugug: This isn't a promise, so there is no catch, and it's all manipulation of things in our system so not gonna get any errors

Snugug:
@scottnath 

* None of Punchcard's functionality is documented in the README. I'm not sure if adding API documentation there makes sense, but we do eventually need to document Punchcard's functionality
* Updated, take a look

scottnath:
@Snugug failed test:
```
apis › APIs: Types
            operator: ===
            expected: [object Object],[object Object],[object Object],[object Object],[object Object]
            actual: [object Object],[object Object],[object Object],[object Object],[object Object]
```

Snugug:
This is the test that sometimes and sometimes does not fail. I may need to re-write this test

Snugug:
@scottnath I frickin give up. logging out the things that need to be identical makes the tests pass, but not having them logged out makes it fail. So ya know what, I'm keeping the logging in

PR #173, opened 2016-07-28
Add dynamic robots.txt

Snugug:
Probably don't want CMSes to be crawled, or maybe you do. This will dynamically place the correct `robots.txt` file in place

---


Snugug:
@scottnath A quick one, but needed if these will live on the public internet

Issue #159, opened 2016-07-28
Developer: Update Revision Page to Reflect New Flow


poofichu:
_As a_ developer _I want_ to update the revision page _so that_ a designer can go in and style the page to mimic a form page.

```gherkin
Feature: Update revision page
  As a developer
  I want to update the revision page
  So that a designer is able to get in and style the page to reflect the same layout as a content additions form.

  Scenario: a user visits the revision page
    Given a common design/layout
    When a user visits the revision page
    Then they will be able to make revision decisions easier.
```

---

scottnath:
PR: https://github.com/punchcard-cms/punchcard/pull/175

For the first one, we had discussed using css and keeping the existing HTML to create the styles the same as the form - but not go through the system to create a form with `disabled` elements.

Can you use the existing HTML with proper classes to do this styling?

poofichu:
YES.  DL's are pretty easy to manipulate.  I'll work my magic.

PR #175, opened 2016-08-01
detritus removed from approval

scottnath:
Extra data removed from approval page

---


scottnath:
closes #159

PR #210, opened 2016-08-05
fixes complatibility with content-types

ayush2k:
Adds blank array to the config to make it search for the input plugins

`DCO 1.1 Signed-off-by: {{Ayush Gupta}} <{{ayush2k@gmail.com}}>`

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -21,6 +21,9 @@ module.exports = {
       revisions: 'revisions',
       save: 'save',
     },
+    plugins: {
+      directory: [],
+    },
     directory: path.join(__dirname, '../content-types'),
     messages: {
       content: {


config/default.js
```diff
@@ -21,6 +21,9 @@ module.exports = {
       revisions: 'revisions',
       save: 'save',
     },
+    plugins: {
```
Snugug: Why do we need to add this in?
In Content Types, [if this doesn't exist, an empty object is passed in](https://github.com/punchcard-cms/content-types/blob/master/lib/content-types/merge.js#L10-L22) to plugabilly. If `content-types` breaks if this isn't available, that should be fixed there. If it's broken in plugabilly, should be fixed there

scottnath: agreed. plugabilly expects `config.search` to be an array

So that means in content types we must provide an array. And it should check if what it sends to plugabilly is an array first

Snugug: Plugabilly should only care if it's an array [if `config.search` is there](https://github.com/punchcard-cms/plugabilly/blob/df72dada47e754777937c81fbd4c5c804335a0cd/lib/meta.js#L16), so we should only need Content Types to ensure that it is passing an array if it wants to pass something along

scottnath: Talking with Ayush:

1. check if `content.plugins.directory` exists
-no? make configPlugins.search a blank array
2. if exists, is it an array?
-yes? send to plugabilly
-no? do 3.
3. if not array, is it a string?
-yes? put in array, send to pb
-no? error out?

@ayush2k sound right?
@Snugug sound legal?

Snugug: Yah sounds about right to me

ayush2k:
@scottnath @Snugug Closing this. 
This issue is solved by 
https://github.com/punchcard-cms/content-types/pull/81

PR #211, opened 2016-08-05
Fixes body structure for session

ayush2k:
Calls format when the session is it

`DCO 1.1 Signed-off-by: {{Ayush Gupta}} <{{ayush2k@gmail.com}}>`

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -156,39 +156,53 @@ const config = values => {
  */
 const format = body => {
   const data = {};
-  Object.keys(body).forEach(key => {
-    if ((key === 'sunset-date') || (key === 'sunset-time') || (key === 'sunrise-date') || (key === 'sunrise-time')) {
-      data[key] = {
-        'value': body[key],
-      };
-    }
-    else {
-      const inputs = key.split('--');
-      const inputName = inputs[0];
-      const inputType = inputs[1];
+  const single = Object.keys(body).filter(input => {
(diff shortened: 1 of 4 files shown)


lib/utils.js
```diff
@@ -156,39 +156,53 @@ const config = values => {
  */
 const format = body => {
   const data = {};
-  Object.keys(body).forEach(key => {
-    if ((key === 'sunset-date') || (key === 'sunset-time') || (key === 'sunrise-date') || (key === 'sunrise-time')) {
-      data[key] = {
-        'value': body[key],
-      };
+  const single = Object.keys(body).filter(input => {
+    if ((input === 'sunset-date') || (input === 'sunset-time') || (input === 'sunrise-date') || (input === 'sunrise-time')) {
```
scottnath: @ayush2k if this is in place, can it supersede [this section in workflows](https://github.com/punchcard-cms/punchcard/blob/master/lib/routes/workflows.js#L93)?

Meaning...can that section in workflows.js be removed?

```
          // clear extraneous values
          const remove = ['sunrise-date', 'sunrise-time', 'sunset-time', 'sunset-date', 'submit', 'action', 'comment-textarea'];
          remove.forEach(rmv => {
            delete data[rmv];
          });
```

note please that there are two extra items in that one: `action` & `comment-textarea`

ayush2k: @scottnath No I don't think so. Because format is still keeping these values

Issue #137, opened 2016-07-23
Developer: Content name is not displaying in content list


joshkimmell:
As a content creator, I want to be able to see a list of content by name, so that I know what content to choose to "edit".

**I don't see the names**
![image](https://cloud.githubusercontent.com/assets/8161298/17074265/6570aaa8-503f-11e6-9b7f-3e074a288cfe.png)

scottnath:
@joshkimmell this is being worked on.

There is a story for tracking:
https://github.com/punchcard-cms/punchcard/issues/143

and a PR with some open questions:
https://github.com/punchcard-cms/content-types/pull/72

Once that PR is complete, I'll update Punchcard using the new `key` and that should fix this issue.

scottnath:
This is complete.

PR #215, opened 2016-08-09
Feature/identifier

### LINKED ISSUE #61: Primary attribute
Integrate identifier into Punchcard.

```gherkin
Feature: Use identifier
  As a core developer
  I want to use a content type's identifier as the title for content
  So that I have a way to reference the content in management pages

  Scenario: View a content type list
    Given that content types are shown in UX as a list of content
      And that content-type identifiers are not-yet required
    When viewing content in lists
    Then I need each piece of content to have a title
      And I need that title to be the content-type's identifier
      And I need a text-placeholder for content without data inside the identifier.

  Scenario: View content approval
    Given that a piece of content may be integrated into an approval workflow
      And that content-type identifiers are not-yet required
    When viewing content
    Then I need that content to have a title
      And I need that title to be the content-type's identifier
      And I need a text-placeholder for content without data inside the identifier.
```

---

- [ ] Use identifier / View a content type list
- [ ] Use identifier / View content approval

scottnath:
Adding identifier to punchcard templates
---
Resolves #61, #137

### DIFF
--- a/lib/routes/utils.js
+++ b/lib/routes/utils.js
@@ -18,7 +18,7 @@ const identifier = (rows, type) => {
     row.identifier = row.value[type.identifier][input].value;
 
     if (!row.identifier) {
-      row.identifier = '(No identifier available)';
+      row.identifier = `Revision: ${row.revision}`;
     }
 
     return row;
--- a/lib/routes/utils.js
+++ b/lib/routes/utils.js
@@ -15,9 +15,12 @@ const identifier = (rows, type) => {
 
   return rows.map(rw => {
     const row = rw;
-    row.identifier = row.value[type.identifier][input].value;
+    const value = _.get(row, `value[${type.identifier}][${input}].value`, '');
 
-    if (!row.identifier) {
+    if (value !== '') {
+      row.identifier = value;
+    }
+    else {
       row.identifier = `Revision: ${row.revision}`;
     }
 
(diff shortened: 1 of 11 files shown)


lib/routes/utils.js
```diff
@@ -0,0 +1,30 @@
...
+const identifier = (rows, type) => {
+  const attr = _.find(type.attributes, { 'id': type.identifier });
+  const input = Object.keys(attr.inputs)[0];
+
+  return rows.map(rw => {
+    const row = rw;
+    row.identifier = row.value[type.identifier][input].value;
+
+    if (!row.identifier) {
+      row.identifier = '(No identifier available)';
```
Snugug: Instead of this, can we use `Revision ${row.id}`?

tests/server.js
```diff
@@ -51,7 +51,14 @@ const service = [
     language: 'test-dummy-entry',
     publishable: false,
     approval: 2,
-    value: {},
+    value: {
+      'service-name': {
+        'text':
```
Snugug: The children of `service-name` shouldn't need to be quoted

Snugug: Although I think that our rule is "consistent"

tests/utils.js
```diff
@@ -159,3 +159,68 @@ test('Request Format', t => {
...
+        'service-name': {
+          'text': {
+            'value': 'This is the test title',
+          },
+        },
+      },
+    },
+  ];
+  const type = {
+    'identifier': 'service-name',
```
Snugug: shouldn't need to quote these

PR #273, opened 2016-08-16
Feature/reference

### LINKED ISSUE #45: Reference Input Plugin
Need an input plugin for referencing one piece of content (with fields) from another.

```gherkin
Feature: Reference Content
     As a Content Creator
     I want to reference one piece of content from another
     So that those pieces of content can be considered together

  Scenario: Reference
    Given that I want to reference content-type A while writing content in content-type B
     When I am writing a piece of content in content-type B
     Then I can reference a piece of content in content type A by selecting it from a list
      And that list can either be a select box or radio buttons
      And the content is represented by its `identifier` attribute of the latest revision
      And the content is referenced by its UUID
```

---
**Reference Content**
- [ ] Reference

ayush2k:
Content to Content Reference

---
Resolves #45 

`DCO 1.1 Signed-off-by: Ayush Gupta <ayush2k@gmail.com>`


ayush2k:
@Snugug @scottnath

lib/routes/content.js
```diff
@@ -9,11 +9,9 @@ const content = require('punchcard-content-types');
 const multipart = require('connect-multiparty');
 const uuid = require('uuid');
 const _ = require('lodash');
-
 const utils = require('../utils');
```
scottnath: We've been informally adding this space between npm dependencies and local `require`s. @Snugug should this be a formal decision to keep this space?

Snugug: Yah, it should be. Should find or write an ESLint plugin for this

config/default.js
```diff
@@ -22,6 +22,9 @@ module.exports = {
       save: 'save',
       new: 'Save Revision',
     },
+    plugins: {
+      directory: [path.join(__dirname, '../input-plugins')],
```
scottnath: There should be unit tests to make sure this works when punchcard is used as a dependency.

lib/routes/content.js
```diff
@@ -85,38 +84,39 @@ const routes = application => {
      * @param {object} next - Express callback
     */
     app.get(`/${config.content.base}/:type/${config.content.actions.add}`, (req, res, next) => {
-      const type = utils.singleItem('id', req.params.type.toLowerCase(), types);
-      const workflow = workflows.workflow(type, allflows, config, req);
-      const errors = _.get(req.session, `form.content.add[${req.params.type.toLowerCase()}].errors`, {});
-      const values = _.get(req.session, `form.content.add[${req.params.type.toLowerCase()}].content`, {});
+      utils.fill(types, req.params.type.toLowerCase(), references, database).then(type => {
```
scottnath: I missed the reason of change here - what changed from populating reference in content-types/merge?
My assumption is it's to have the most up-to-date references to select from

ayush2k: We don't have access to database in content-types. Secondly, the content in content-types can change while the server is running and merge will not be able to reflect those changes as it is called on initialization.

config/default.js
ayush2k: Need a change. I don't think it will work with the dependency.

input-plugins/input-plugin-reference/lib/validation.js
```diff
@@ -0,0 +1,25 @@
...
+ * @param {Object} [settings.all.reference] - settings of input reference {Object}
+ *
+ * @returns {Bool|String} true or a string with the text describing the error
+ *
+ * @module referenceValidation
+ */
+
+module.exports = function referenceValidation(input, settings) {
+
+  return true;
```
Snugug: This should be an API call to check to ensure that the selected item exists, although we may not need to do this w/this PR

lib/routes/content.js
Snugug: The conversation was to do it as part of the merge step in the routing, but never in `content-types` itself

lib/routes/content.js
```diff
@@ -231,108 +233,109 @@ const routes = application => {
...
-        return next();
-      }
+        const steps = (JSON.parse(JSON.stringify(workflow.steps))).reverse();
 
-      const steps = (JSON.parse(JSON.stringify(workflow.steps))).reverse();
+        // new revision, re-start workflow process
+        const step = steps[steps.length - 1];
 
-      // new revision, re-start workflow process
-      const step = steps[steps.length - 1];
```
Snugug: Where did this go?

ayush2k: It is still there.

content-types/bar.yml
```diff
@@ -0,0 +1,31 @@
+name: Bar
```
scottnath: This file only should be in ./tests/content-types please

input-plugins/input-plugin-reference/index.js
```diff
@@ -0,0 +1,39 @@
+
+'use strict';
+
+/**
+ * checkbox Input Plugin
```
scottnath: Fix file comments.

input-plugins/input-plugin-reference/index.js
```diff
@@ -0,0 +1,39 @@
...
+ *
+ *
+ * An input plugin with multiple checkboxes
+ */
+const validation = require('./lib/validation.js');
+
+/**
+ * Single checkbox Input Plugin
```
scottnath: comments

input-plugins/input-plugin-reference/lib/validation.js
scottnath: 👍 
however, agree on 2nd point too, not this pr please.

lib/utils.js
```diff
@@ -229,6 +230,147 @@ const format = body => {
...
+      Object.keys(inputs).forEach(input => {
+        // Checks if input is a reference
+        if (inputs[input].hasOwnProperty('reference') && inputs[input].reference === true) {
+          // Throw error if contentType is not defined
+          if (!inputs[input].settings.hasOwnProperty('contentType')) {
+            throw new Error('Reference must have a content type');
+          }
+
+          // Throw error if contentType is not valid
+          if (!position.hasOwnProperty(inputs[input].settings.contentType)) {
```
scottnath: could there be more error checking here?
* is it a string
* is it empty

ayush2k: It will throw an error if the content type is not available.
So if there is any invalid string or empty content-type, it will throw an error

lib/utils.js
```diff
@@ -229,6 +230,147 @@ const format = body => {
...
+        const options = [];
+        const map = {};
+
+        // Gets the lates revison and pushes unique values to options
+        rws.forEach(row => {
+          if (!map.hasOwnProperty(row.id)) {
+            map[row.id] = true;
+
+            // Checks if data is stored in the correct structure
+            if (row.value.hasOwnProperty(types[input.ct.index].identifier)) {
```
scottnath: We should make the `identifier` attribute automatically `require: save`. Then we wouldn't need to have this check.

ayush2k: I didn't want this check but it doesn't work for the old values stored in the database without this.
I think I should remove this check

lib/utils.js
```diff
@@ -229,6 +230,147 @@ const format = body => {
...
+  const ref = refs.filter(reference => {
+    return ct.id === reference.type;
+  });
+
+  return Promise.map(ref, input => {
+    // Get the records from content-type
+    return database
+      .select('*')
+      .from(`content-type--${input.ct.id}`)
+      .orderBy('revision', 'DESC').then(rws => {
```
scottnath: I think you could use [unique](http://knexjs.org/#Chainable-unique) in this database query and have it only find one of each piece of content. Is this the correct criteria?

* unique id
* order by approved asc (0=approved; lower numbers further along process)
* order by created

lib/utils.js
```diff
@@ -229,6 +230,147 @@ const format = body => {
...
+        const options = [];
+        const map = {};
+
+        // Gets the lates revison and pushes unique values to options
+        rws.forEach(row => {
+          if (!map.hasOwnProperty(row.id)) {
```
scottnath: is `row.id` the same as `row.revision`?

lib/utils.js
```diff
@@ -229,6 +230,147 @@ const format = body => {
...
+            map[row.id] = true;
+
+            // Checks if data is stored in the correct structure
+            if (row.value.hasOwnProperty(types[input.ct.index].identifier)) {
+              options.push({
+                value: row.id,
+                label: row.value[types[input.ct.index].identifier][Object.keys(row.value[types[input.ct.index].identifier])[0]].value,
```
scottnath: You can run the array of rows that come out of the database through the [routes/utils/identifier function](https://github.com/punchcard-cms/punchcard/blob/master/lib/routes/utils.js#L12) to give each row their identifier - which ends up at `row.identifier` - which would be cleaner to use.

lib/utils.js
scottnath: https://github.com/punchcard-cms/content-types/issues/83

scottnath:
I've looked over the whole thing. Great work. I think we could clean up a chunk of the logic if we moved up https://github.com/punchcard-cms/content-types/issues/83.

scottnath:
made identifier required to save: https://github.com/punchcard-cms/content-types/pull/94

lib/routes/content.js
scottnath: this promise should move down and be chained with the .only promise below.

lib/routes/workflows.js
```diff
@@ -32,108 +33,109 @@ const routes = application => {
...
-
-      // check workflow exists
-      if (!workflow) {
-        return next();
-      }
-
-      const steps = (JSON.parse(JSON.stringify(workflow.steps))).reverse();
-
-      _.unset(req.session, 'form.content.approve');
+      utils.fill(types, req.params.type.toLowerCase(), references, database).then(type => {
```
scottnath: this promise should move down and be chained with the database promise below.

lib/utils.js
ayush2k: Neither unique nor distinct is working

ayush2k: changed the query. Now working

lib/utils.js
ayush2k: done

lib/routes/workflows.js
ayush2k: this promise gives us the type. And without type, we will be able to call `workflow` and `check.type`

content-types/bar.yml
ayush2k: when testing using agent, we get content-types from punchcard content-types. With just one CT, I will not be able to test reference

ayush2k:
@scottnath @Snugug updated

config/default.js
scottnath: What was the fix for this?

lib/routes/workflows.js
scottnath: actually, we could stick with the original way to get `type`, (`utils.singleItem...`) then, for `utils.fill`, you could send in `type` instead of `types`. That would clean up some error checking on fill and remove the need for `ct.index`

lib/utils.js
```diff
@@ -229,6 +229,144 @@ const format = body => {
...
+ * @param {object} types - content types object
+ * @param {string} id - id of the requested content type
+ * @param {array} refs - reference data
+ * @param {object} database - database object
+ *
+ *
+ * @returns {object} - content type object with reference data
+ */
+const fill = (types, id, refs, database) => {
+  const ct = singleItem('id', id, types);
```
scottnath: sending in `type` would remove the need for this.

lib/routes/workflows.js
scottnath: (I may be wrong about `ct.index`...I admit I'm confused by the usage of that)

lib/utils.js
```diff
@@ -229,6 +229,144 @@ const format = body => {
...
+
+    // Get the records from content-type
+    return database
+      .select('id', 'revision', 'value')
+      .from(`content-type--${input.ct.id}`)
+      .whereIn('revision', where).then(rws => {
+        const rows = routes.identifier(rws, types[input.ct.index]);
+        const options = rows.map(row => {
+          return {
+            value: row.id,
```
scottnath: can this be `row.revision`?

lib/utils.js
ayush2k: we were calling `singleItem` to get the type.
So, we can either call this function before `fill` or here.
Will not make any difference

lib/utils.js
ayush2k: In that case, we will be referencing a revision instead of a content

lib/utils.js
ayush2k: done

lib/utils.js
ayush2k: no `row.id` is the id of the content

lib/utils.js
ayush2k: done

lib/routes/workflows.js
ayush2k: Should I move the promise and chain it only? It would work but the value of type will be outdated during the call to workflow and check.type.

lib/utils.js
scottnath: OH. Duh, thank you for the clarification.

tests/utils.js
```diff
@@ -294,30 +294,93 @@ test('Reference array has values', t => {
...
+          index: 1,
+          id: 'test-service',
+        },
+      },
+    ],
+  };
   const result = utils.references(types);
-  t.is(result.length, expected.length, 'length of arrays is equal');
-  t.is(JSON.stringify(result, null, 2), JSON.stringify(expected, null, 2), 'Reference array created');
+  console.log(JSON.stringify(result, null, 2));
```
scottnath: remove please

config/default.js
ayush2k: We need another way to tell content-types to grab all the input-plugin from the host `punchcard`. In the future, we might ship more plugins with punchcard.

tests/utils.js
ayush2k: done

lib/routes/content.js
```diff
@@ -321,23 +323,23 @@ const routes = application => {
...
-            return content.only(type.id, data, [type]);
+            return content.only(ct.id, data, [ct]);
           }).then(only => {
             return content.form(only);
           }).then(form => {
             res.render('content/add', {
               form,
               action: `/${config.content.base}/${req.params.type.toLowerCase()}/${config.content.actions.save}`,
-              type,
+              ct,
```
scottnath: then key needs to stay `type` or the view will break.

ayush2k: I will change it. But how is it working

ayush2k: done

lib/routes/content.js
```diff
@@ -85,38 +86,40 @@ const routes = application => {
      * @param {object} next - Express callback
     */
     app.get(`/${config.content.base}/:type/${config.content.actions.add}`, (req, res, next) => {
-      const type = utils.singleItem('id', req.params.type.toLowerCase(), types);
-      const workflow = workflows.workflow(type, allflows, config, req);
-      const errors = _.get(req.session, `form.content.add[${req.params.type.toLowerCase()}].errors`, {});
-      const values = _.get(req.session, `form.content.add[${req.params.type.toLowerCase()}].content`, {});
+      const ct = utils.singleItem('id', req.params.type.toLowerCase(), types);
+      utils.fill(types, ct, references, database).then(type => {
```
scottnath: @ayush2k this needs to be moved down to chain with the other promises as well.

ayush2k: done

ayush2k: Done

config/default.js
Snugug: Why doesn't this work? What is the value of `plugins.directory` in an implementation that's pulling in the default configuration from Punchcard? Can we force a value here in `init`?

ayush2k: content-types uses globalconfig to get the plugins and I cannot think of a way to change globalconfig on init

lib/utils.js
```diff
@@ -229,6 +229,146 @@ const format = body => {
...
+ *
+ * @param {object} types - content types object
+ * @param {object} type - content type
+ * @param {array} refs - reference data
+ * @param {object} database - database object
+ *
+ *
+ * @returns {object} - content type object with reference data
+ */
+const fill = (types, type, refs, database) => {
```
Snugug: Instead of passing in the database, this module should require our database and use that required instance

ayush2k: I tried it but database comes out to be a blank object

lib/utils.js
```diff
@@ -238,4 +378,6 @@ module.exports = {
   time,
   config,
   format,
+  fill,
```
Snugug: As these do not have general functionality across all files, they should be in `content/utils`

lib/utils.js
```diff
@@ -229,6 +229,146 @@ const format = body => {
...
+ * @returns {object} - content type object with reference data
+ */
+const fill = (types, type, refs, database) => {
+  const ct = _.cloneDeep(type);
+  if (!ct) {
+    return new Promise(resolve => {
+      resolve(ct);
+    });
+  }
+  const ref = refs.filter(reference => {
```
Snugug: New line here please

lib/utils.js
ayush2k: fill and references uses routes/utils and also `singleItem`. And we cannot create dependency of utils in content/utils.

lib/utils.js
```diff
@@ -229,6 +229,146 @@ const format = body => {
...
+    return ct.id === reference.type;
+  });
+
+  return Promise.map(ref, input => {
+    // subquery for select
+    const where = database
```
Snugug: this looks like your swallowed promise. Does `wherIn` in like 349 take a promise or a function? Or both?

ayush2k: Both but function use this and cannot be an arrow function

lib/utils.js
```diff
@@ -229,6 +229,146 @@ const format = body => {
...
+          if (!position.hasOwnProperty(inputs[input].settings.contentType)) {
+            throw new Error(`Content Type ${inputs[input].settings.contentType} is not valid`);
+          }
+
+          if (_.get(inputs[input], 'settings.view') === 'radio') {
+            inputs[input].type = 'radio';
+          }
+
+          // Defines and push reference object
+          refs.push({
```
Snugug: you're using `map` here, use that to return your array.

Snugug: `cts.map`, you should return the map of `Object.keys`

lib/utils.js
```diff
@@ -229,6 +229,146 @@ const format = body => {
...
+          .max('revision')
+          .from(`content-type--${input.ct}`)
+          .groupBy('id');
+
+    // Get the records from content-type
+    return database
+      .select('id', 'revision', 'value')
+      .from(`content-type--${input.ct}`)
+      .whereIn('revision', where).then(rws => {
+        const rows = routes.identifier(rws, singleItem('id', input.ct, types));
```
Snugug: `singleItem` should be able to be replaced with [`Array.find`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) to reduce dependency on the actual `utils` file. This should also probably be done before the function call in order to reduce our call stack as it's not dependent on the result of the function it's in

Snugug:
* Please add a test for the `return new Promise` in `utils.fill`
* In content type edit route, there are [no tests](https://coveralls.io/builds/7494212/source?filename=lib%2Froutes%2Fcontent.js#L283) for your new work. Please add some

lib/utils.js
Snugug: You can write a normal function, just need to include its parameter name. That, however, doesn't answer the question of does `whereIn` take a promise or a function? If it only takes a function, not a promise, that's likely the swallowed promise

lib/utils.js
Snugug: Do you know why?

ayush2k: No

lib/utils.js
ayush2k: I think just a function

lib/utils.js
```diff
@@ -229,6 +229,150 @@ const format = body => {
...
+ * @param {object} types - content types object
+ * @param {object} type - content type
+ * @param {array} refs - reference data
+ * @param {object} database - database object
+ *
+ *
+ * @returns {object} - content type object with reference data
+ */
+const fill = (types, type, refs, database) => {
+  const ct = _.cloneDeep(type);
```
Snugug: There's no error checking here to ensure that `type` is something consumable (specifically if it's not all lowercase) like there is where this is called

Snugug:
Merged! Thanks for being awesome this summer @ayush2k! The remaining issues we'll work out later

PR #282, opened 2016-08-17
Hotfix/e2e refactor

scottnath:
Removing routes from e2e testing. Round 1.

---
Resolves (incomplete on content)


Snugug:
@scottnath CI fails, timeout

Snugug:
@scottnath Correct me if I'm wrong, but one of the goals of this is to reduce reliance on `server.js` and being able to hit our endpoints for code coverage, right? If so, I'd be curious to see what our code coverage looks like without any route coverage whatsoever

Snugug:
I fear that code coverage may still be hidden as if you hit the route you hit the functions you're refactoring, and we have the same problem, just with better structured code

Snugug:
I'd almost be interested in splitting our routing tests out of or main test scaffolding and running them separately w/o code coverage

scottnath:
A big chunk of what we run into without spinning up punchcard as we do in server.js is all of the init files get missed. I want server.js to eventually just be a route-checker (does it get 404s in all instances needed, do routes work otherwise)

I'm creating a database story this morning for setting up a testing db (or some other option) for unit testing these routes further.

scottnath:
With server.js:
```
All files                                            |    81.86 |    71.06 |        0 |    82.28 |                |
```

Without server.js:
```
All files                                            |    39.48 |    54.02 |        0 |    39.68
```

With only one test of the top-level url in server.js:
```
All files                                            |    64.73 |    57.88 |        0 |    65.06 |                |
```

scottnath:
Before refactor:

```
All files |    34.42 |    45.27 |        0 |    34.61
```

Snugug:
@scottnath Yah, so `server.js` is hiding about 40% code coverage that we need to get in to unit tests. I'd like to see it removed all-together.

Snugug:
Basically our E2E tests are hiding the fact that an additional 40% of our code isn't covered, just because it gets hit when spinning up a route

lib/content/routes.js
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * @returns {promise} database promise returned
+ */
+const one = (res, type) => {
+  return database
+    .distinct(database.raw('ON (id) id'))
+    .select('*')
+    .from(`content-type--${type.id}`)
+    .orderBy('id', 'DESC')
+    .orderBy('revision', 'DESC').then(rws => {
```
Snugug: Check out this query Ayush and I came up with yesterday. It's used in enough places (this makes two) that we should probably abstract it out:

```SQL
select * from "content-type--services" where
revision in (select max("revision") from "content-type--services" group by "id") 
```
It selects all fields from content types where the revision is the highest revision, grouped by unique IDs. This will just return the most recent revisions of each piece of content.

lib/content/routes.js
```diff
@@ -0,0 +1,93 @@
...
+      // add itentifier to each row
+      const rows = utils.routes.identifier(rws, type);
```
Snugug: I'm thinking we should update the database model to have a separate field for identifier so we don't need to run identifier over every DB call to get the identifier

lib/content/routes.js
```diff
@@ -0,0 +1,93 @@
...
+const database = require('../database');
+const utils = require('../utils');
+
+/**
+ * All Content-Types Landing page
+ * @param {object} res - HTTP Response
+ * @param {object} allTypes all content types available to the CMS
+ */
+const all = (res, allTypes) => {
+  res.render('content/home', {
```
Snugug: The other functions in this file all `return res.render` from a promise. Why isn't this being returned? Should we continue our practice from init of using promises as a shared interface even if they aren't necessarily required and turn this in to a promise?

lib/content/routes.js
```diff
@@ -0,0 +1,93 @@
...
+        });
+      });
+    });
+  });
+};
+
+module.exports = {
+  all,
+  one,
+  add,
```
Snugug: Why isn't the `save` route stuff in here?

lib/content/routes.js
scottnath: Sounds like a great story!
Not sure if this comment is more than pointing out a cool thing...

lib/content/routes.js
scottnath: Sounds great...out of scope for this PR. Please create a story

lib/content/routes.js
scottnath: because I was cut off when working on this and did not finish.

lib/content/utils.js
```diff
@@ -1,74 +1,126 @@
 'use strict';
```
Snugug: Are these changes mostly around GitHub being wonky about showing diff?

lib/content/routes.js
Snugug: No, it's "this should be the query being used" more than "cool query"

scottnath:
We can't remove a file like server.js all together. we still need to check _each_ route for positive and negative factors. 
But, we should figure out how we're going to test all the `init` code - which is actually a pretty big hole in terms of coverage and out-of-scope for the routes refactor.

lib/content/routes.js
Snugug: I would like all of content moved over before merging this; especially `save` as that's the one I need this story merged for

Snugug:
I'd live to remove `server.js` from our code-coverage covered unit tests so checking to see that they exist/fail as expected doesn't hide our code coverage

Snugug:
That's literally the point of the E2E refactor

lib/content/routes.js
scottnath: I don't think we should continue that practice here. Those init scripts all build on each other in a row. These routes are stand-alone in respect to each other.

res.render is not typically returned (https://expressjs.com/en/guide/routing.html). I think it's appropriate in promises, but not if there is not promise.

scottnath:
If I remove it, then the numbers for code coverage will tank and this PR will not be able to be merged. _And_ the actual routes will not be checked that they work - which _does_ require a file like server.js. We can unit test the internals of routes, but not the `app.get` for routes - which must be done by spinning up express

lib/content/routes.js
scottnath: Totally. 1. Still a cool query. 2. story still needs to be written

lib/content/utils.js
```diff
@@ -1,74 +1,126 @@
...
+ * @type object
+ *
+ * @property {object} type - Content type configuration, after merged with input plugins
+ * @property {object} workflow - workflow configuration
+ * @property {string} id - id of a piece of content
+ * @property {string} revision - revision number for a revision of a piece of content
+ */
+
+/**
+ * Checking that what is in a content url matches what's in the system
```
Snugug: This description needs to be changed. My assumption was that it was going to return true/false, not that it was going to return an object gleaned from the URL.

Also,  not sure if I like that it's returning a string with an error as opposed to causing an error of some sort. This feels like a good use of a promise as an interface for us, where we either resolve with the object or reject with an error string instead of a sync function that a user will need to have checks in place themselves to determine if it's passed or failed

lib/content/utils.js
```diff
@@ -1,74 +1,126 @@
...
 
 const config = require('config');
-const _ = require('lodash');
 const isUUID = require('validator/lib/isUUID');
 
+const workflows = require('../workflows/utils');
 
-const check = {
+/**
+ * Grab type from all content types in system
```
Snugug: For all of these, using promises as a shared interface feels like the right approach so a user doesn't need to have their own checking in place to determine if what's been returned is a valid value or not.

lib/content/utils.js
```diff
@@ -1,74 +1,126 @@
...
+  // url has a revision in it
+  if (req.params.revision) {
+    if (typeof revision(req.params.revision) === 'string') {
+      return revision(req.params.revision);
+    }
+  }
+
+  return {
+    type: typ,
+    workflow,
```
Snugug: I'd like to see all assumed attributes like this at the bottom of an object

lib/content/utils.js
```diff
@@ -1,74 +1,126 @@
...
+    id: req.params.id,
+    revision: req.params.revision,
+  };
 };
 
 module.exports = {
-  check,
+  check: {
```
Snugug: Why are you namespacing all of this behind `check`?

lib/routes/content.js
```diff
@@ -53,29 +48,20 @@ const routes = application => {
      * @param {object} next - Express callback
     */
     app.get(`/${config.content.base}/:type`, (req, res, next) => {
-      const type = utils.singleItem('id', req.params.type.toLowerCase(), types);
+      const good = utils.content.check.url(req, types, allFlows);
```
Snugug: Seeing this here, I would really love to see this as a promise, especially given that `routes` is a promise; all of the errors could roll up nicely

lib/routes/content.js
```diff
@@ -86,41 +72,18 @@ const routes = application => {
...
-              type: ct,
-              config: config.content,
-              step,
-            });
-          });
-        }).catch(e => {
-          next(e);
-        });
-      });
+      return routes.add(req, res, good.type, good.workflow, references, types);
```
Snugug: As we control the signature for this function, why not just have it take `found` instead of `type` and `workflow` and have `routes.add` pick out `type` and `workflow` for itself?

lib/content/routes.js
Snugug: https://github.com/punchcard-cms/punchcard/issues/288

lib/content/routes.js
Snugug: https://github.com/punchcard-cms/punchcard/issues/288

lib/content/routes.js
Snugug: We discussed that `save` was harder to abstract out and will likely need to wait for another PR

lib/routes/content.js
```diff
@@ -234,33 +194,25 @@ const routes = application => {
...
-      }
+      const type = good.type;
+      const workflow = good.workflow;
 
-      // check workflow exists
-      if (!workflow) {
-        return next();
-      }
+      const errors = _.get(req.session, `form.content.add[${req.params.type.toLowerCase()}].errors`, {});
+      const idSess = _.get(req.session, `form.content.add[${req.params.type.toLowerCase()}].id`, {});
```
Snugug: `sessID`, `sessRev`

lib/workflows/utils.js
```diff
@@ -68,6 +69,37 @@ const check = (config) => {
...
+    });
+  }
+  else {
+    wf = allFlows.find(flow => {
+      return flow.id === globalConfig.workflows.default;
+    });
+  }
+
+  if (wf === undefined || typeof wf !== 'object') {
+    return globalConfig.workflows.messages.missing.replace('%type', type.name).replace('%workflow', type.workflow);
```
Snugug: Same comment I have here as I have in `content/utils`, I'd prefer to use promise-as-an-interface here and not return a string for error or object for true and have to  have the places that implement this need to check for that each time

tests/content-routes.js
```diff
@@ -0,0 +1,153 @@
...
+    t.end();
+  });
+  response.render();
+});
+
+//////////////////////////////
+// Routes - One content type's landing route
+//////////////////////////////
+test.cb('Working single content landing route', t => {
+  const response = httpMocks.createResponse({ eventEmitter: EventEmitter });
```
Snugug: Is `node-mocks-http` a replacement for supertest? If so, why are we using this instead of supertest?

scottnath:
@Snugug ... decided to revisit the `url` checks here and make them middleware instead. Much more appropriate and easier to maintain/control in the long run. Will work up a PR today.

scottnath:
@nugug https://github.com/punchcard-cms/punchcard/issues/305

scottnath:
@Snugug closing this so I can create a new PR once #309 gets merged.

This PR was too clouded by the error checking of the content routes. Once that error checking is in place, this PR will be drastically cleaner.

PR #287, opened 2016-08-18
Make checkboxes and radios accessible/tabbable

### LINKED ISSUE #275: Bug: A11Y Problem with Checkbox Styling
When tabbing through a form, checkboxes (and presumably radios) don't tab. This is due to them having **[display](https://github.com/punchcard-cms/punchcard/blob/57f173f9c4476b2c0a13f7a21439f3020d950634/src/sass/partials/global/_mixins.scss#L54-L55): [none](https://github.com/punchcard-cms/punchcard/blob/57f173f9c4476b2c0a13f7a21439f3020d950634/src/sass/partials/global/_mixins.scss#L130-L131)** set. These need to be visually hidden, not removed entirely.

## Steps for Reproducing

- Go to a content type with checkboxes
- Try and tab to a checkbox
- Notice it skipped

## Screenshots

Can't Tab to Checkboxes

## Affected Browsers

- [ ] Chrome
- [ ] Firefox

(presumably all others)

## Runtime Version

v5.11

## Code Version

Latest

poofichu:
Resolves #275


scottnath:
👍 
defer to @Snugug

Snugug:
@poofichu Your branch needs to be updated

src/sass/partials/global/_mixins.scss
```diff
@@ -62,11 +62,11 @@
       border: setting-get('input border width') solid setting-get('primary color');
       content: '';
       display: inline-block;
-      height: 1em;
-      margin-right: setting-get('rhythm') / 2;
-      transform: translateY(-.125em);
+      height: 1.25em;
+      margin-right: setting-get('rhythm') / 4;
+      transform: translateX(-.8rem);
```
Snugug: * You're mixing `rem` and `em` for translation in `checkbox`. You should use one or the other
* Where does the `.8` come from? That seems like a slightly arbitrary number

src/sass/partials/global/_mixins.scss
```diff
@@ -78,6 +78,17 @@
...
     }
   }
+
+  &:focus {
+    outline: 0 none;
+
+    + label {
+
+      &::before {
+        outline: 4px solid setting-get('primary accent color');
```
Snugug: Is this the same outline we have throughout Punchcard? If so, we should probably turn it in to a variable to better reuse it

src/sass/partials/global/_mixins.scss
```diff
@@ -139,17 +150,27 @@
       border-radius: 50%;
       content: '';
       display: inline-block;
-      height: 1em;
-      margin-right: setting-get('rhythm') / 2;
-      transform: translateY(-.125em);
+      height: 1.25em;
+      transform: translateX(-.9rem) translateY(-.125em);
```
Snugug: Again mixing `em` and `rem` for translate

src/sass/partials/global/_mixins.scss
```diff
@@ -139,17 +150,27 @@
...
       vertical-align: middle;
-      width: 1em;
+      width: 1.25em;
     }
   }
 
   &:checked + label {
     &::before {
-      background: radial-gradient(color('grey', 80) 40%, transparent 40%);
+      background: radial-gradient(setting-get('primary color') 50%, setting-get('white') 50%);
```
Snugug: Where is this mixin coming from?

src/sass/partials/global/_mixins.scss
poofichu: I was using `rem` to match up with what we're doing with txt/font but I can switch it to `em`

poofichu: Yes, .8 is arbitrary.  I'll find something that works with our `rhythm` setting as is divisable by all the things.

src/sass/partials/global/_mixins.scss
poofichu: Grabbed the code from our original design guide.  Looks like it's coming from a node module.

Snugug: Can you please just write the standard Radial Gradient? That way we can can have [autoprefixer](https://github.com/postcss/autoprefixer) do its job

Snugug:
w/r/t your emoji usage, as this was resolving a bug, it's not new functionality (:new: `:new:`), it's a bug fix (:bug: `:bug:`)

poofichu:
Noted about the bug emoji.  Habit of doing 🆕

src/sass/partials/global/_mixins.scss
Snugug: Oh, wow, NEVER MIND

Snugug: There's no `@include`. This is fine

poofichu:
Even a @Snugug can make a mistake every now and then 😛

Now MERGE PLZZZZZZZZZ

src/sass/partials/global/_mixins.scss
```diff
@@ -52,7 +52,7 @@
 }
 
 @mixin checkbox {
-  display: none;
+  transform: translateX(.5em);
```
Snugug: This doesn't entirely hide the radio buttons. I tried with `translateX(.25em) translateY(-.15em)` and that put the radio element entirely behind the custom styling. With `.5em` it peeks out the bottom right, which is very prominent on iOS

**.5em**
screen shot 2016-09-01 at 1 48 08 pm
screen shot 2016-09-01 at 1 48 03 pm

**.25em .15em**
screen shot 2016-09-01 at 1 48 31 pm
screen shot 2016-09-01 at 1 48 15 pm

src/sass/partials/global/_mixins.scss
```diff
@@ -128,7 +138,7 @@
 }
 
 @mixin radio {
-  display: none;
+  transform: translateX(.5em);
```
Snugug: This doesn't entirely hide the radio buttons. I tried with `translateX(.25em) translateY(-.15em)` and that put the radio element entirely behind the custom styling. With `.5em` it peeks out the bottom right, which is very prominent on iOS

**.5em**
screen shot 2016-09-01 at 1 48 08 pm
screen shot 2016-09-01 at 1 48 03 pm

**.25em .15em**
screen shot 2016-09-01 at 1 48 31 pm
screen shot 2016-09-01 at 1 48 15 pm

Snugug:
@poofichu Out of date again

PR #309, opened 2016-08-24
Hotfix/route middlware

### LINKED ISSUE #305: URL checking for content types
A primary hurdle to the refactor of our code is having a common way to check that urls have what is needed to move forward to show a page to the user. Content-type management pages have a specific URL path pattern which can be tested.

```gherkin
Feature: URL checks before page load
  As a core developer
  I want content management urls to be confirmed for correctness

  Scenario: URL
    Given that routes in Punchcard must follow specific patterns
      And that those routes can be broken up into testable parts-of-the-path
     When setting up routes in Punchcard
     Then I need to have middleware that checks each part of a path for correctness
      And it should check a content path's parts start with an existing content type
      And it should check a content id is in UUID format
      And it should check a content revision is a number
      And it should check the url's action exists among available actions

  Scenario: Workflow
    Given that content management pages require a workflow
    When testing routes for content in Punchcard
    Then I need to have the system check that a content-type's workflow exists

  Scenario: Error
    Given that a user needs to know what has gone wrong with their attempt to visit a url
    When a content route path test fails
    Then the error should be captured
      And the error should trigger the next step to show the user what has gone wrong.
```

---

**URL Checks**
- [x] URL
- [x] Workflow
- [x] Errors

scottnath:
Adding route checking to content. Required to move forward on refactor story in a clean, clear way.

---
Resolves #305

### DIFF
--- a/config/default-test.js
+++ b/config/default-test.js
@@ -15,7 +15,7 @@ if (process.env.CI === 'true') {
       database: 'travis_ci_test',
     },
     debug: false,
-    acquireConnectionTimeout: 1000,
+    acquireConnectionTimeout: 2000,
   };
 }
 
--- a/config/default.js
+++ b/config/default.js
@@ -117,7 +117,7 @@ module.exports = {
   workflows: {
     messages: {
       approved: 'This content has already completed the approval process',
-      missing: 'Workflow \'%workf\' for Content Type \'%type\' not found',
+      missing: 'Workflow \'%workflow\' for Content Type \'%type\' not found',
     },
   },
 };
--- a/config/default.js
+++ b/config/default.js
@@ -52,7 +52,7 @@ module.exports = {
       database: 'punchcard',
     },
     debug: false,
-    acquireConnectionTimeout: 1000,
+    acquireConnectionTimeout: 2000,
   },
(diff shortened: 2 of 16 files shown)


Snugug:
If you were to split this up in to multiple pull requests, how could you split this up?

scottnath:
Well, first commit is a separate repair - to get errors properly working - but only a few lines.

Then there could be a second PR of just fixtures - which are 312 of this 759 line this PR (41%). I could def see a seperate PR of that as there are a number of fixtures that need to be re-jiggered.

Then the workflow utils changes could be another one - although because that is now a promise instead, it would take more work to integrate the promise of that separately, but doable as it has it's own set of tests.

The new content utils could be a standalone pr. Then the integration of it could be another PR. Although, that would make /lib/content/utils.js weird because it would have to retain all the old code along with all the new stuff - which is very very different. So...those need to share a PR. Or be in different files.

Snugug:
Sounds like a good plan! Go for it!

scottnath:
@Snugug #313, #316

scottnath:
closing...makin a whole buncha small ones!

PR #313, opened 2016-08-25
errors not caught with custom err handler

### LINKED ISSUE #312: Error route refactor
Errors are not ending up in our custom error handler. 

```gherkin
Feature: Error handling through routes
  As a core developer
  I want to have errors use our custom route error handler
  So that I can control how our error pages work

  Scenario: Convert error routes
    Given that some code can be broken out of routes
     When I need to fix error routing
     Then I should create new code and tests to cover the repaired route code.

  Scenario: Cascade error route changes
    Given that errors are going to be triggered in a different way
    When I fix the error route for generic errors
    Then I need to fix any error triggers in existing routes to use the updated routes
```

---

**Unit Testing**
- [x] Convert
- [x] Cascade

scottnath:
This covers a bug which makes our app ignore the error route we created. Instead it uses Express' built in error route page currently.

Also upgrading error routes to be more test-able! What fun!

---
Resolves #312

### DIFF
--- a/config/default-test.js
+++ b/config/default-test.js
@@ -15,7 +15,7 @@ if (process.env.CI === 'true') {
       database: 'travis_ci_test',
     },
     debug: false,
-    acquireConnectionTimeout: 1000,
+    acquireConnectionTimeout: 2000,
   };
 }
 
--- a/lib/errors/routes.js
+++ b/lib/errors/routes.js
@@ -66,7 +66,7 @@ const errors = (err, req, res, next, app) => {
 
     res.status(error.status || 500);
 
-    if (error.status === '404') {
+    if (error.status === 404) {
       res.render('404', error);
     }
     else {
--- a/lib/errors/routes.js
+++ b/lib/errors/routes.js
@@ -15,7 +15,7 @@ const _ = require('lodash');
  *
  * @returns {promise} promise containing the route
  */
-const notFound = (req, res) => {
+const missing = (req, res) => {
   return new Promise((resolve) => {
     const message = _.get(req.session, '404.message', `Not Found: ${req.url}`);
     const safe = _.get(req.session, '404.safe');
@@ -65,12 +65,19 @@ const errors = (err, req, res, next, app) => {
     }
 
     res.status(error.status || 500);
-    res.render('error', error);
+
+    if (error.status === '404') {
+      res.render('404', error);
+    }
+    else {
+      res.render('error', error);
+    }
+
     resolve(true);
   });
 };
 
 module.exports = {
-  notFound,
+  missing,
   errors,
 };
--- a/lib/errors/routes.js
+++ b/lib/errors/routes.js
@@ -1,0 +1,76 @@
+/**
+ *  @fileoverview Error routes
+ *
+ *  @author  Scott Nath
+ *
+ */
+'use strict';
+
+const _ = require('lodash');
+
+/**
+ * Route for a 404 error
+ * @param {object} req - HTTP Request
+ * @param {object} res - HTTP Response
+ *
+ * @returns {promise} promise containing the route
+ */
+const notFound = (req, res) => {
+  return new Promise((resolve) => {
+    const message = _.get(req.session, '404.message', `Not Found: ${req.url}`);
+    const safe = _.get(req.session, '404.safe');
+
+    _.unset(req.session, '404');
+
+    res.status(404);
+    res.render('404', {
+      message,
+      safe,
+    });
+    resolve(true);
+  });
+};
+
+/**
+ * Route for errors
+ *
+ * @param {string|object} err - error object or error message string
+ * @param {object} req - HTTP Request
+ * @param {object} res - HTTP Response
+ * @param {object} next - Express callback
+ * @param {object} app Express application
+ *
+ * @returns {promise} promise containing the route
+ */
+const errors = (err, req, res, next, app) => {
+  return new Promise((resolve) => {
+    let error = {
+      message: 'route error',
+      status: '',
+      safe: '/',
+    };
+
+    if (typeof err === 'object') {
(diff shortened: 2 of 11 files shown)


lib/errors/routes.js
```diff
@@ -0,0 +1,76 @@
...
+      error.message = err;
+    }
+
+    error.error = {};
+
+    // If in development, will print stacktrace
+    if (app.get('env') === 'development') {
+      error.error = err;
+    }
+
```
Snugug: If `status===404`, use our 404 render

lib/routes/errors.js
```diff
@@ -21,16 +21,7 @@ module.exports = application => {
...
-      const safe = _.get(req.session, '404.safe');
-
-      _.unset(req.session, '404');
-
-      res.status(404);
-      res.render('404', {
-        message,
-        safe,
-      });
+      return routes.notFound(req, res);
```
Snugug: Can we call this `routes.missing` instead of `notFound` just so we can avoid `camelCase`?

lib/routes/users.js
```diff
@@ -111,7 +111,8 @@ const routes = application => {
         if (rows.length < 1) {
           res.status(404);
```
Snugug: Instead of setting status here, can we build an error object and have our error handler deal with this? This will help ensure we don't accidentally set our status multiple times.

tests/errors.js
```diff
@@ -0,0 +1,139 @@
+import test from 'ava';
+import events from 'events';
+import httpMocks from 'node-mocks-http';
```
Snugug: why not `mocks`?

tests/errors.js
```diff
@@ -0,0 +1,139 @@
...
+import _ from 'lodash';
+
+import routes from '../lib/errors/routes';
+
+const EventEmitter = events.EventEmitter;
+
+// used in a promise below
+let nextCalled = false; // eslint-disable-line no-unused-vars
```
Snugug: why not `called`

Snugug: Why are you setting this up outside the function call? Why can't this live inside the `next` mock?

tests/errors.js
```diff
@@ -0,0 +1,139 @@
...
+  get,
+};
+
+/**
+ * Next
+ * @param  {object} err - the error object
+ * @returns {object} nextCalled - the error object
+ */
+const next = (err) => {
+  nextCalled = err;
```
Snugug: Here, for instance (https://github.com/punchcard-cms/punchcard/pull/313/files#r76660294)

tests/errors.js
Snugug: or `http`

tests/errors.js
```diff
@@ -0,0 +1,139 @@
...
+//////////////////////////////
+// Routes - 404
+//////////////////////////////
+test.cb('404 error route', t => {
+  const request = httpMocks.createRequest({
+    method: 'GET',
+    url: '/some/madeup/icky/path',
+  });
+
+  request._setSessionVariable('404', {});
```
Snugug: Does this need to be two different lines?

tests/errors.js
```diff
@@ -0,0 +1,139 @@
...
+
+  _.set(request.session, '404', {
+    message: 'Something bad happened',
+    safe: '/refreshing/safe/path',
+  });
+
+  const response = httpMocks.createResponse({ eventEmitter: EventEmitter });
```
Snugug: Can we keep our `const`s together if they don't have internal dependencies on each other?

tests/errors.js
```diff
@@ -0,0 +1,139 @@
...
+  routes.notFound(request, response);
```
Snugug: Or at the ver least some space between the two and a comment setting up the actual request/response?

tests/server.js
```diff
@@ -755,7 +755,7 @@ test.cb('Users Edit Landing Page', t => {
     });
 });
 
-test.cb('Users Delete Landing Page', t => {
+test.skip('Users Delete Landing Page', t => {
```
Snugug: Why are we skipping this test in this PR?

views/error.html
```diff
@@ -1,8 +1,11 @@
-<h1 class="base--h1">I'm error!</h1>
+<div class="not-found">
```
Snugug: Why did you add an `error.html` view instead of re-using the `404` view directly?

Snugug:
@scottnath You've got comments, and where they apply to multiple places in your code, I'd like to see those updated as well

views/error.html
scottnath: Because I wasn't sure of the reasoning of a separate 404 vs error view. Although both using the _exact same_ template now, I wanted to leave 404 on it's own so that any changes to a 404 page in the future would not affect the general error page.

Snugug: ok

lib/errors/routes.js
```diff
@@ -0,0 +1,83 @@
...
+    res.status(error.status || 500);
+
+    if (error.status === '404') {
```
Snugug: Is this going to come in as a string or a number? If it comes in as a number, this is going to fail, because `===`

Snugug: It looks like you're setting them as numbers FWIW

Snugug: You should also include tests to check to make sure the right error handler is being hit; I would have expected this to fail your tests

scottnath:
@Snugug changes added. waiting on CI.

config/default-test.js
```diff
@@ -15,7 +15,7 @@ if (process.env.CI === 'true') {
       database: 'travis_ci_test',
     },
     debug: false,
-    acquireConnectionTimeout: 1000,
```
Snugug: With #323 merged, this needs to be removed/branch updated

PR #316, opened 2016-08-25
Hotfix/test fixtures

### LINKED ISSUE #315: Test fixtures
Fixtures for tests in punchcard have been hastily slapdashed in a dogpile of nope. We need them to use a more common directory structure and old/outdated files removed.

```gherkin
Feature: Test Fixtures cleanup
  As a core developer
  I want a clean, easy-to-understand directory containing my tests fixtures
  So that I can have a quick way to mimic Punchcard and content-types output.

  Scenario: Common directory
    Given that fixtures currently reside in a haphazard set of folders under `/tests`
      And all fixtures should be under the `/tests/fixtures` directory
     When traversing through Punchcards tests directory
     Then all fixtures should be under the `/tests/fixtures` directory

  Scenario: New common fixtures
    Given that I need common fixtures across test files
    When updating or creating new tests
    Then I should have some common fixtures which represent the needs of various Punchcard functions
      And there should be a fixture object for multiple content types after yaml import, after merging with input plugins
      And there should be a fixture object for multiple workflows after yaml import.
```

---

**Test fixtures**
- [x] Common Directory
- [x] New common fixtures

scottnath:
Fixtures have been re-arranged and now the tests properly point to the actual test fixtures.

---
Resolves #315


config/default-test.js
```diff
@@ -1,9 +1,12 @@
 'use strict';
 
+const _ = require('lodash');
 const path = require('path');
 
 const defaultCfg = require('./default');
 
+const config = _.cloneDeep(defaultCfg);
```
Snugug: Why are you cloning this? The `config` module will deep merge these together, so we should only need to override the specific object we want to change

lib/workflows/load.js
```diff
@@ -32,6 +32,7 @@ const load = (wfpath) => {
     dir.readFiles(workflowsPath, {
       match: /.yml$/,
       exclude: /^\./,
+      recursive: false,
```
Snugug: Why are you turning off recursive find?

package.json
```diff
@@ -20,8 +20,8 @@
     "posttest": "gulp build",
     "start": "nodemon -e html,js ./index.js",
     "lint": "gulp lint",
-    "ava:watch": "NODE_CONFIG_DIR=../config ava --watch | tap-diff",
-    "ava": "NODE_CONFIG_DIR=../config ava | tap-diff",
+    "ava:watch": "NODE_ENV=default-test NODE_CONFIG_DIR=../config ava --watch | tap-diff",
```
Snugug: This _should_ just be "test" and maybe we rename the test configuration to go with it

tests/workflows.js
```diff
@@ -380,5 +342,5 @@ test('workflows in type', t => {
   noflow.workflow = '';
   const nopeflow = workflows.workflow(noflow, allFlows, globalConfig, req);
 
-  t.is(nopeflow, false, 'Returns false on workflow missing from global flows');
+  t.is(nopeflow, allFlows[1], 'Returns false on workflow missing from global flows');
```
Snugug: Using array index feels brittle. Thoughts?

tests/fixtures/content-types/all-types-merged.js
```diff
@@ -0,0 +1,281 @@
+/**
```
Snugug: Either `all-types` or `all-merged` but not `all-types-merged` please.

config/default-test.js
scottnath: I see. So in this case, I need to make a change to `config.content` so I should cloneDeep just that object you're saying?

lib/workflows/load.js
scottnath: will remove

package.json
scottnath: agreed....so then `./config/test.js` ..._but_ is that the file which travis will look for?

PR #330, opened 2016-08-30
workflow promise

### LINKED ISSUE #329: Get workflow via promise
I need to update the acquisition of workflows in preparation for a route middleware.

```gherkin
Feature: Workflow from workflows
  As a core developer
  I want a content-type's workflow
  So that I can use the steps for various functions

  Scenario: Content-type
    Given my content type requires a workflow
      And that my workflow should be among the workflow configurations available to me
     When I grab data about my content type
     Then I need to attain the workflow id from my content type's configuration
      And I need to attain the full workflow configuration from all workflows in my app
      And I need to return an error if no matching config is found.

  Scenario: Default workflow
    Given my content type requires a workflow
      And that my content type configuration does not have a workflow id in it
     When I grab data about my content type
     Then I need to attain the default workflow id from my app's configuration
      And I need to attain the default workflow configuration using the default workflow id
      And I need to return an error if no matching config is found.
```

---

**Workflow from workflows**
- [x] Content type
- [x] Default

scottnath:
Creating new workflow-grabbing function with a promise for upcoming route middleware

---
Resolves #329

### DIFF
--- a/lib/workflows/utils.js
+++ b/lib/workflows/utils.js
@@ -91,7 +91,7 @@ const workflow = (type, workflows) => {
       });
     }
 
-    if (wf === undefined || typeof wf !== 'object') {
+    if (typeof wf !== 'object') {
       rej(globalConfig.workflows.messages.missing.replace('%type', type.name).replace('%workflow', type.workflow));
     }
 
--- a/lib/workflows/utils.js
+++ b/lib/workflows/utils.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const globalConfig = require('config');
 const slugify = require('underscore.string/slugify');
 
 /*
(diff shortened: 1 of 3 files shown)


lib/workflows/utils.js
```diff
@@ -68,6 +69,37 @@ const check = (config) => {
...
+        return flow.id === type.workflow;
+      });
+    }
+    else {
+      wf = workflows.find(flow => {
+        return flow.id === globalConfig.workflows.default;
+      });
+    }
+
+    if (wf === undefined || typeof wf !== 'object') {
```
Snugug: This just needs to be `typeof wf !== 'object`. If `wf` is undefined, it'll be `typeof undefined`

Snugug:
One comment, otherwise GTG

scottnath:
@Snugug fix added. Awaiting tests.

PR #332, opened 2016-08-30
url parameter checking

### LINKED ISSUE #331: Check url parts
I need to test various parts of a url to confirm they are indeed the correct type of information. 

```gherkin
Feature: URL in request
  As a core developer
  I want a request's URL to be checked for errors
  So that I can avoid badly-formatted urls throwing unknown errors.

  Scenario: type
    Given I have a request object with a url in it
      And that it should be a url with multiple parameters
      And that one of those parameters is a content type's id
      And I have an object containing all content types associated with my application
     When testing the content type's id parameter 
     Then I need to test that the id is a string
      And I need to test that the id matches an id of an associated content type
      And I need to receive the matching content type object
      Or I need to receive an error message.

  Scenario: id
    Given I have a request object with a url in it
      And that it should be a url with multiple parameters
      And that one of those parameters is a piece of content's id
     When testing the parameters
     Then I need to test that the id is a string
      And I need to test that the id is in uuid format
      And I need to receive a confirmation that it is a uuid string
      Or I need to receive an error message.

  Scenario: revision
    Given I have a request object with a url in it
      And that it should be a url with multiple parameters
      And that one of those parameters is a revision number
     When testing the parameters
     Then I need to test that the revision is a string
      And I need to test that the revision is a number
      And I need to receive a confirmation that it is a number and a string
      Or I need to receive an error message.

  Scenario: url
    Given I have a request object with a url in it
      And that it should be a url with multiple parameters
      And that each of those parameters must match certain formats and types
     When receiving a request
     Then I need to test the parameters of the url
      And if those parameters all pass I need the request object changed
      And I need that change to include the full object of the content type referenced by the url
      Or I need to receive an error message.
```

---

**URL in request**
- [x] type
- [x] id
- [x] revision
- [x] url

scottnath:
@Snugug 

Adding in new url parameter checking. Co-exists in this PR with existing parameter checking because this is the foundation for a new route middleware. These new checks cannot supersede the existing checks because it would require a _large_ amount of re-writing to incorporate the promises - which would _all_ be throw-away code as this is all bound for a middleware function.

---
Resolves #331


lib/content/utils.js
```diff
@@ -69,6 +69,118 @@ const check = {
...
+/**
+ * Grab type from all content types in system
+ *
+ * @param {object} req - Express Request Object
+ *
+ * @returns {promise} reject with error message; resolve with single content-type object
+ */
+const type = (req) => {
+  return new Promise((res, rej) => {
+    const typ = req.params.type.toString();
```
Snugug: Need to check to see if `type` exists first, otherwise this will 'splode

lib/content/utils.js
```diff
@@ -69,6 +69,118 @@ const check = {
...
+
+    if (!typ || typ === '') {
+      rej(config.content.messages.missing.type.replace('%type', '(no type given)'));
+    }
+
+    const exists = req.content.types.find(tp => {
+      return tp.id === typ.toLowerCase();
+    });
+
+    if (!exists || (exists === false) || (typeof exists !== 'object') || !exists.id || !exists.name) {
```
Snugug: This feels… ugly. Can this if statement be cleaned up? For instance, `typeof exists !== 'object` will catch `!exists` and `exists === false`

lib/content/utils.js
```diff
@@ -69,6 +69,118 @@ const check = {
...
+ * Determine if an id is in uuid format
+ *
+ * @param {object} req - Express Request Object
+ *
+ * @returns {promise} reject with error message; resolve true
+ */
+const id = (req) => {
+  return new Promise((res, rej) => {
+    // no id in the request
+    if (!req.params.hasOwnProperty('id')) {
```
Snugug: If it doesn't have an ID, its in UUID format?

lib/content/utils.js
```diff
@@ -69,6 +69,118 @@ const check = {
...
+      res(true);
+    }
+
+    // this can be the add, approve, or save action instead of a uuid
+    if (req.params.id === config.content.actions.add || req.params.id === config.content.actions.save || req.params.id === config.content.actions.approve) {
```
Snugug: When would the `id` parameter (that is dynamic) be one of our hardcoded path items? Also, these are "true" for UUIDs?

lib/content/utils.js
```diff
@@ -69,6 +69,118 @@ const check = {
...
+
+/**
+ * Checking that what is in a content url matches what's in the system
+ * @param {object} req - HTTP Request
+ * @param {object} res - HTTP Response
+ * @param {object} next - Express callback
+ *
+ * @returns {function} express next function
+ */
+const url = (req, res, next) => {
```
Snugug: Why is this function signature so drastically different than the other ones in this file? Why does this get `req`, `res`, and `next`?

lib/content/utils.js
scottnath: if it doesn't have an ID, we're done checking this URL. 

The url is set up like this:
`/content/:type/:id/:revision/:action`
if the third section of that url is not present (`id`), then the url is going to go the `/content/:type` if there is anything after `:type` the system would default to that being checked by the `id` section as the system ignores multiple slashes. Meaning this is not a thing:
`/content/:type//:revision/:action`
it would just become this thing:
`/content/:type/:id/:revision`

lib/content/utils.js
scottnath: not a uuid check...specific to these actions which happen on a url that doesn't have an `id` or or `revision` in it. If it is one of those three...then this url is legal. The third tier can be an `id` in uuid format _or_ one of these actions.

lib/content/utils.js
scottnath: This is the pre-cursor to the route middleware. The middleware will give this function those parameters and only this function needs `next`. This function doesn't use `res`, but because this is middleware the order of these is already set.
The other functions do not need `next` as they are not triggering next but instead just taking their specific input and checking it for accuracy - it's this function which will decide what to do with the accuracy determination.

scottnath:
@Snugug function/file locations have changed as per our conversation.

Any other changes needed?

lib/content/middleware/index.js
```diff
@@ -0,0 +1,47 @@
...
+const url = (req, res, next) => {
+  const request = req;
+  const err = {
+    message: '',
+    safe: `/${config.content.base}`,
+    status: 404,
+  };
+
+  // if there are no url parameters or type parameter, we're done checking this url
+  if (!request.params || !request.params.hasOwnProperty('type')) {
```
Snugug: `typeof request.params === undefined`

lib/content/middleware/utils.js
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * @param {object} req - Express Request Object
+ *
+ * @returns {promise} reject with error message; resolve with single content-type object
+ */
+const type = (req) => {
+  return new Promise((res, rej) => {
+    const typ = req.params.type.toString();
+
+    if (!typ || typ === '') {
```
Snugug: `typeof type === 'undefined'`

lib/content/middleware/utils.js
```diff
@@ -0,0 +1,86 @@
...
+      rej(config.content.messages.missing.type.replace('%type', '(no type given)'));
+    }
+
+    const exists = req.content.types.find(tp => {
+      return tp.id === typ.toLowerCase();
+    });
+
+    if ((typeof exists !== 'object') || !exists.id || !exists.name) {
```
Snugug: `exists.hasOwnProperty('id')`

Snugug: `exists.hasOwnProperty('name')`

lib/content/middleware/utils.js
```diff
@@ -0,0 +1,86 @@
...
+    if (!req.params.hasOwnProperty('id')) {
+      res(true);
+    }
+
+    // this can be the add, approve, or save action instead of a uuid
+    if (req.params.id === config.content.actions.add || req.params.id === config.content.actions.save || req.params.id === config.content.actions.approve) {
+      res(true);
+    }
+
+    if (req.params.id === '' || typeof req.params.id !== 'string' || !isUUID(req.params.id)) {
```
Snugug: Won't `!isUUID` catch if it's empty and not a string?

lib/content/middleware/utils.js
```diff
@@ -0,0 +1,86 @@
...
+ * @returns {promise} reject with error message; resolve true
+ */
+const revision = (req) => {
+  return new Promise((res, rej) => {
+    // no revision in the request
+    if (!req.params.hasOwnProperty('revision')) {
+      res(true);
+    }
+
+    if (req.params.revision === '' || !Number.isInteger(parseInt(req.params.revision, 10))) {
```
Snugug: Won't the Number check catch if it's an empty string?

lib/content/middleware/utils.js
scottnath: could remove the empty part, have to keep string

lib/content/middleware/utils.js
scottnath: it does!

lib/content/middleware/utils.js
scottnath: fixed

lib/content/middleware/utils.js
scottnath: fixed

lib/content/middleware/index.js
scottnath: fixed

scottnath:
@Snugug gtg

Issue #335, opened 2016-08-31
Bug: Labels Deploy


Snugug:
Label deploy is failing with a 127 error. As this is non-essential and causing our tests to error out, we should remove the deploy.

https://travis-ci.org/punchcard-cms/punchcard/builds/156532368#L731

## Version/Node

Latest/5.11

PR #350, opened 2016-09-02
Deploy labels, again

Snugug:
An attempt to actually get #335 working

---


Snugug:
@scottnath I believe this will allow us to start deploying label changes.

PR #359, opened 2016-09-07
Hotfix/ct config

### LINKED ISSUE #339: Content-types configuration
A module we use for core functionality is changing the way it accepts configuration. Instead of an app-based globally-reached config, it now requires a config as a parameter. We need to change our application to provide a configuration as a parameter to this module.

```gherkin
Feature: Module requires config as parameter
  As a core developer
  I want to control the configuration provided from Punchcard to the `content-types` module
  So that I can dynamically configure said module based on one or multiple config sources

  Scenario: Standalone app
    Given that my application uses the `content-types` module
      And that I want to provide my core input plugin path to `content-types`
    When I use content types in my application
    Then I should provide a configuration to `content-types`
      And the path to my core input plugin directory should be included in said configuration..

  Scenario: App as module
    Given that my application uses the `content-types` module
      And that my application can be used as an included module in another app
      And that I provide core configuration for my application
      And that users may optionally provide my application with configuration
    When I my application is being used as a module
    Then I should perform a merge of a user's config with my core config
      And said config should include a path to my application's core input plugins directory
      And said config may optionally include a path or paths to an implementor-provided input plugins directory or directories.
```

---

**Module accepts config parameter**
- [x] Standalone app
- [x] App as module

scottnath:
Adds init function to combine app/punchcard plugin directory configs

---
Resolves #339 

### DIFF
--- a/lib/init/config.js
+++ b/lib/init/config.js
@@ -1,0 +1,38 @@
+'use strict';
+
+const config = require('config');
+const _ = require('lodash');
+const path = require('path');
+
+const local = path.join(__dirname, '../../input-plugins');
+
+module.exports = (application) => {
+  return new Promise(res => {
+    const app = application;
+
+    // grab plugins dir from config
+    let plugins = _.get(config, 'content.plugins.directory', []);
+
+    // convert string to array
+    if (typeof plugins === 'string') {
+      plugins = [plugins];
+    }
+
+    // must be string or array
+    if (!Array.isArray(plugins)) {
+      throw new Error('Plugin directory configuration must be string or an array');
+    }
+
+    // push local directory if it does not exist
+    if (plugins.indexOf(local) === -1) {
+      plugins.push(local);
+    }
+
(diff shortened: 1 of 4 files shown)


lib/init/config.js
```diff
@@ -0,0 +1,38 @@
...
+
+    // push local directory if it does not exist
+    if (plugins.indexOf(local) === -1) {
+      plugins.push(local);
+    }
+
+    const configuration = _.set(config, 'content.plugins.directory', plugins);
+
+    // use node-config function to add Punchcard plugin directory
+    config.util.extendDeep(config, configuration);
```
Snugug: Can we actually guarantee that this will always be able to change the `config` object before it becomes immutable?

scottnath: the only way we could run into an immutable config or partial config is if the implementer used [makeImmutable](https://github.com/lorenwest/node-config/wiki/Using-Config-Utilities#makeimmutableobject-propertyname-propertyvalue). If they do that, really, what can we do?

PR #360, opened 2016-09-07
Feature/content middleware

### LINKED ISSUE #314: Content Route Middleware
Our content routes all have common checks that need to be done on the various parameters that make up a content's url. These checks should be done within an express middleware.

```gherkin
Feature: URL checks before route creation
  As a core developer
  I want content management urls to be confirmed for correctness before the route-specific code runs

  Scenario: Middleware
    Given that our system can perform url-parts checks before it triggers a route's code
    When a route path is checked for correctness
    Then that test should happen as middleware
      And that middleware should be triggered for every content route
      And it should trigger an error route on url errors
      And it should continue to create the route with zero url errors.

  Scenario: Return data
    Given that the middleware can add data to the original request
    When a route path has been tested
      And the tests come back with no errors
    Then the middleware should return route-required data
      And that data should include the content-type's merged object
      And that data should include the content-type's workflow object.

  Scenario: Redirect to error page
    Given that a user has attempted to visit a url which is incorrectly formed
      And that the user needs to know what mistake has been made
    When a content route's url creates an error
    Then the user should receive an error message
      And the message should be a path-part specific error message for the problem
      And it should show the user an error page
      And that error page should contain the path-part specific error message
      And that error page should contain a safe url for the user to visit to try again.
```

---

**URL checks**
- [x] Middleware
- [x] Return data
- [x] Redirect to error page

scottnath:
Adds the content middleware and uses the variables added to `request` in the routes.

---
Resolves #314


Snugug:
@scottnath branch out of date. otherwise looks good

Snugug:
Resolves #362 
Resolves #347

PR #369, opened 2016-09-09
Feature/rise set approver views

### LINKED ISSUE #354: Sunrise/Sunset Approver Views
Split approver scenarios from #136 out

**Definitions**
* Now - The current date and time
* Live - Publicly available in the API and the `live` database the API draws from
* Sunrise - The date, time, and timezone content is scheduled to go live
* Sunset - The date, time, and timezone content is schedule to stop being live
* Approved- Content that has gone through all steps in its [workflow](https://github.com/punchcard-cms/punchcard/pull/105)

```gherkin
Feature: Approve Content
  As a content approver
  I want to understand how the content I'm reviewing is going to be affected by sunrise/sunset
  So that I can make an informed decision about whether or not I should approve or reject the revision being reviewed

  Scenario: Cancel Sunset
    Given that I am reviewing a content revision without a sunset
     When that revision's content has an existing live revision
      And that the live content has a sunset
      And the sunset has not yet occurred
     Then I should be notified on the approval page that approving this revision will cancel the existing sunset
      And I should be notified of the existing sunset

  Scenario: Change Sunset
    Given that I am reviewing a content revision with a sunset
      And that revision's sunset is past now
     When that revision's content has an existing live revision
      And that the live content has a sunset
      And the sunset has not yet occurred
     Then I should be notified on the approval page that this revision will change the existing sunset
      And I should be notified of the extent of the change (how much further/closer to now the change is)

  Scenario: Immediate Sunset
    Given that I am reviewing a content revision with a sunset
      And that revision's sunset is before now
     When that revision's content has an existing live revision
      And that the live content has a sunset
      And the sunset has not yet occurred
     Then I should be notified on the approval page that this revision will immediately sunset its content 

  Scenario: Sunrise That Which Is Sunset
    Given that I am reviewing a content revision with a sunrise
     When that revision's content has an existing live revision
      And that the live content has a sunset
      And the sunset has already occurred
     Then I should be notified on the approval page that this revision will sunrise the currently sunset content

  Scenario: Sunrise That Which Will Be Sunset
    Given that I am reviewing a content revision with a sunrise
     When that revision's content has an existing live revision
      And that the live content has a sunset
      And the sunset has not yet occurred
      And the sunset will occur before the current revision's sunrise
     Then I should be notified on the approval page that this revision will sunrise the content that otherwise would be sunset
      And I should be notified of the difference between when the content would be scheduled to sunset and sunrise

  Scenario: Immediate Sunrise
    Given that I am reviewing a content revision with a sunrise
     When that revision's sunrise is equal to or before now
     Then I should be notified on the approval page that this revision will be live immediately after approval

  Scenario: Sunrise Override
    Given that I am reviewing a content revision with a sunrise
     When that revision's content has existing revisions scheduled for sunrise
      And this revision's sunrise is the same as other revisions that are scheduled for sunrise
     Then I should be notified on the approval page that this revision will supersede the other scheduled revisions with the same sunrise

  Scenario: View Sunrise and Sunset
    Given that I am reviewing a content revision
     When I want to check the sunrise or sunset value of the revision being reviewed
     Then I should be able to readily see them just as I can the other fields of a revision
```

---

**Approve Content**
- [x] Cancel Sunset
- [x] Change Sunset
- [x] Immediate Sunset
- [ ] ~~Sunrise That Which Is Sunset~~
- [x] Sunrise That Which Will Be Sunset
- [x] Immediate Sunrise
- [x] Sunrise Override
- [x] View Sunrise and Sunset

Snugug:
Adds Approver Sunrise/Sunset Views

---
Resolves #354 


lib/utils.js
```diff
@@ -84,18 +84,19 @@ const isoTime = (date, time, zone) => {
 const inputTime = (date, zone) => {
   let tz = zone;
 
-  if (date === '' || date === undefined) {
+  if (date === '' || date === undefined || date === null) {
     return null;
   }
 
-  if (typeof tz === 'undefined') {
+  if (tz === undefined || tz === null) {
```
scottnath: why not `typeof` on this `undefined`?

lib/utils.js
```diff
@@ -84,18 +84,19 @@ const isoTime = (date, time, zone) => {
 const inputTime = (date, zone) => {
   let tz = zone;
 
-  if (date === '' || date === undefined) {
+  if (date === '' || date === undefined || date === null) {
```
scottnath: typeof undefined

lib/utils.js
Snugug: Yah because it doesn't work with stuff coming out of the DB for some reason. They're typeof object

lib/utils.js
Snugug: Yah because it doesn't work with stuff coming out of the DB for some reason. They're typeof object

scottnath:
@Snugug done with comments up to c3e3190

Snugug:
@scottnath I've split the schedule view out to #373, that will need to be done later

Issue #364, opened 2016-09-08
Bug: reference plugin as require


scottnath:
References is not available as a `require` when creating multi-input plugins. It works when using within content-types, but when trying to test a multi plugin using it, the `require` statement fails in node.

## Steps for Reproducing

- Create a multi-input plugin using references in an app that uses Punchcard as a module
- Set up the multi-input to use the `shared-tests`
- Run `npm test`
- Error returned: `Error: Error: Input 'method-example-request' not found`

scottnath:
@Snugug I'm opening this bug so that we can figure out the options to fix this and write stories in kind.

The issue is that when we use this in another plugin:
`const reference = require('input-plugin-reference');`
then node hates it and cannot find it and thinks it's dumb.

## Options:

1. Export the plugins inside punchcard and use them something like this:
`const reference = require('punchcard-cms/plugins/input-plugin-reference');

    _problem_: would *only* work when creating a CMS with Punchcard as a module, but would not work for any multi-input plugins which would be distributed via npm

2. Move reference out of punchard so it can be included as an npm module - and no longer put core plugins in Punchcard's repo

    _problem_: Core modules would probably not work as a standalone module on their own - reference certainly does not

Snugug:
May I introduce you to [peerDependencies](https://docs.npmjs.com/files/package.json#peerdependencies)

scottnath:
So...you vote `2.` - remove from punchcard repo. Correct?

scottnath:
This will now eat my day manana: https://github.com/punchcard-cms/punchcard/issues/365

scottnath:
option #1 is actually the answer. All fixed, closing this issue.

PR #372, opened 2016-09-09
Hotfix/reference remove

### LINKED ISSUE #365: Reference stand alone module
Reference does not work as a module shipping within Punchcard.

```gherkin
Feature: Reference in CMS
  As an implementor
  I want references to be available to use in my forms
  So that I can use references in my CMS.

  Scenario: Reference module
    Given reference needs to be accessible as an npm module
     When I want to use the reference input plugin
     Then I should be able to include it as a dependency in my module

  Scenario: Remove reference
    Given that reference is a plugin inside Punchcard
      And that I need to extricate it from Punchcard and make a standalone module
      And I need Punchcard to work with the reference module
    When I want to use Punchcard to create a CMS
    Then I can create content which references other content
      And I only need to include references as a dependency to make it work
      And I can use Punchcard without reference as a dependency and it does not error.
```

---

**Reference in CMS**
- [x] Reference module
- [x] Remove reference

scottnath:
Reference plugin required to be removed for when Punchcard used as a module

---
Resolves #365 


Snugug:
@scottnath out of date

scottnath:
@Snugug updated and passed tests

Snugug:
Please explain to me why this isn't working as expected before we remove this

scottnath:
The reason this doesn't work is that when we make local input plugins in an app using the reference input plugin, then we cannot use `require` because node cannot find any `require`'s unless they are in node_modules. The only other option is to have an export from punchcard of the plugin - but then that means that no multi-input plugins could be made and shipped via npm unless they use `punchcard` as a dependency. That was what [this bug](https://github.com/punchcard-cms/punchcard/issues/364) was for.

Snugug:
Does this now work? Have you done a full integration test with an instance to confirm this is indeed working as expected?

Snugug:
Everything works as expected currently, with the following bug:

* Config initialization needs to happen before database initialization

That should be how we move forward with reference. Input plugins can require reference by doing `const reference = require('punchcard-cms/input-plugins/input-plugin-reference');`. For input plugins distributed via NPM, they should set punchcard as a peerDependency.

Closing this in favor of this bug fix, then references work.

PR #374, opened 2016-09-09
config switch with database

scottnath:
Quick bugfix

---


scottnath:
@Snugug

PR #387, opened 2016-09-13
Hotfix/api attributes

### LINKED ISSUE #377: Bug: Types Count Wrong
The count of live items in content types is not correct. 

## Steps for Reproducing

- Publish a number of items of the same content type
- Publish a number of items of a different content type
- Notice content count in `meta` doesn't match actual number of published items

## Screenshots

JSON Response

Database

## Affected Browsers

- [ ] All

## Runtime Version

5.11

## Code Version

2.10.0

### LINKED ISSUE #378: Bug: Extra Info in API
API responses contain extra data that they should not.

## Steps for Reproducing

- Publish a piece of content
- Go to the `types` endpoint
  - Observe `workflow` and `identifier` are included but should not be
- Go to its endpoint
  - Observe `sunrise-date`, `sunset-time`, `sunset-date`, `sunset-time`, `submit` items in `attributes`
  - Observe `action` item in `attributes`

## Screenshots

screen shot 2016-09-12 at 10 28 20 am

screen shot 2016-09-12 at 10 27 20 am

screen shot 2016-09-12 at 10 27 15 am

## Affected Browsers

- [ ] All

## Runtime Version

5.11

## Code Version

2.10.0

### LINKED ISSUE #379: Bug: Content Attributes Not Usable
It appears as if the update to saving content makes the APIs response for an individual piece of content unusable and now needs to be updated.

## Steps for Reproducing

- Publish a piece of content
- Go to its endpoint
  - Observe that attributes that have a single value have that nested inside a `value` key which is nested inside its `type` key.
  - Observe that attributes that have multiple values have arrays of objects that have their value inside a `value` key which is nested inside the `type` key.

## Screenshots

screen shot 2016-09-12 at 10 32 01 am

## Affected Browsers

- [ ] All

## Runtime Version

5.11

## Code Version

2.10.0

Snugug:
Fix for APIs being sad

---
Resolves #379 
Resolves #378 
Resolves #377 

### DIFF
--- a/lib/routes/api.js
+++ b/lib/routes/api.js
@@ -15,59 +15,63 @@ const api = require('../api');
 const routes = application => {
   const app = application;
 
-  return api.types(app).then(types => {
-    /*
-     *  API for all Content
-     */
-    // TODO: Tests
-    // TODO: Return error if no items are found
-    app.get('/api', (req, res, next) => {
-      api.all(req.query).then(formatted => {
-        res.json(formatted);
-      }).catch(err => {
-        next(err);
-      });
-    });
+  const types = api.types(app);
 
-    /*
-     *  API for all Content Types
-     */
-    // TODO: Tests
-    // TODO: Return error if no items are found
-    app.get('/api/types', (req, res) => {
-      const formatted = api.content(req.query, types);
-
+  /*
+   *  API for all Content
+   */
+  // TODO: Tests
+  // TODO: Return error if no items are found
+  app.get('/api', (req, res, next) => {
+    api.all(req.query).then(formatted => {
       res.json(formatted);
+    }).catch(err => {
+      next(err);
     });
-
-    /*
-     *  API for all Content of a given Content Type
-     */
-    // TODO: Tests
-    // TODO: Return error if no items are found
-    app.get('/api/types/:type', (req, res, next) => {
-      api.ofType(req.query, req.params.type).then(formatted => {
-        res.json(formatted);
-      }).catch(err => {
-        next(err);
-      });
-    });
-
-    /*
-     *  Specific Piece of Content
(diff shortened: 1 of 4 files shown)


lib/routes/api.js
```diff
@@ -15,59 +15,73 @@ const api = require('../api');
...
     });
+  });
+
+  /*
+   *  Specific Piece of Content
+   */
+  // TODO: Tests
+  // TODO: Return error if no items are found
+  app.get('/api/types/:type/:id', (req, res, next) => {
+    const type = app.get('content-types').find(item => {
```
scottnath: To be uniform, should this be a promise?

lib/routes/api.js
```diff
@@ -15,59 +15,73 @@ const api = require('../api');
...
-      api.one(req.query, req.params.id).then(formatted => {
-        res.json(formatted);
-      }).catch(err => {
-        next(err);
-      });
+      return false;
     });
 
-    return app;
+    if (types.length === 0) {
```
scottnath: perhaps this stuff should be in middleware instead to keep the route as just a promise and to share the test of the url params across routes.

scottnath: spoke with same out-of-scope for this PR

scottnath:
@Snugug only my promise question is [outstanding](http://65.media.tumblr.com/9a7d90beacfee9ad18a9659a36144076/tumblr_inline_nb6udfuWFD1rq5p7o.gif). But may not be because the middleware is not gonna happen here.

Shall I merge then?

lib/routes/api.js
Snugug: No, this is a simple find transform, no need for it to be wrapped in a promise

Issue #328, opened 2016-08-30
Clean Up Lib Folder


Snugug:
When tabbing around in the `lib` folder, having folders and files named identically makes it hard to open. Files should be refactored to `index.js` inside each folder

```gherkin
Feature: Clean Lib
  As a core developer
  I want all files relating to a `lib` directory in that directory
  So that I can easily find and use that directory

  Scenario: Main Export
    Given that our `lib` directories all contain related functionality
      And that the directories are named in a semantic way to indicate what functionality they contain
      And that that functionality is made available together
     When I want to include functionality from that directory
     Then I can `require` that directory to get all related functionality
      And the file ultimately being `require`d is within that directory as well
```

---

**Clean Lib**
- [ ] Main Export

PR #431, opened 2016-09-20
Feature/apps schema

### LINKED ISSUE #430: Applications
Applications are a new type of data being used by and tracked within Punchcard. 

```gherkin
Feature: Applications data structure
  As a core developer
  I want to manage Applications within Punchcard
  So that I can add, edit, and delete Applications where I am using those Applications.

  Scenario: model
    Given that managing Applications requires modeling the structure of the data surrounding them
      And that I can use the established way in which Punchcard can already manage non-content data
      And that there are certain rules for structure required by a Punchcard module, Content-Types
     When setting up Punchcard to manage Applications
     Then I need to create a model for the Application data
      And I need that model to follow the structure required by Content-Types
      And that model will include attributes for each type of data
      And it will include a text attribute for name
      And it will include a text attribute for live endpoint
      And it will include a text attribute for updated endpoint
      And it will include a text attribute for sunset endpoint
      And it will include a text attribute for client id
      And it will include a text attribute for client secret

  Scenario: schema
    Given that Applications will need to be entering into the database
      And to allow that to happen within Punchcard, I need to give Punchcard database schema
    When setting up Applications in Punchcard
    Then I need to create a schema for creating a table for Applications in the database
      And that schema will be an array of objects
      And each object with have a name and type
      And there will be an object for id, which will increment and be the index
      And there will be an object for name, which will be of type string
      And there will be an object for live endpoint, which will be of type string
      And there will be an object for updated endpoint, which will be of type string
      And there will be an object for sunset endpoint, which will be of type string
      And there will be an object for client id, which will be of type string
      And there will be an object for client secret, which will be of type string
      And there will be an object for responses from endpoints, which will be of type jsonb
      And there will be an object for created date, which will be of type timestamp
      And there will be an object for updated date, which will be of type timestamp

  Scenario: fixture
    Given that I will need to test functionality related to Applications in Punchcard
      And that the structure of Applications and it's schema are known
    When setting up tests for Applications functionality
    Then I should have a fixture to mimic that structure
      And it should be a copy of what would be output after the model has been processed by Punchcard's Content-Types module.

  Scenario: Merged
    Given that the Applications model will be processed by the Content-Types module
      And that the Applications model will not change after Punchcard is running
      And that processing a model through Content-Types is processor-heavy
    When I am setting up Applications for initialization in Punchcard
    Then Applications should be processed by Content-Types when Punchcard is initialized
      And I should have the post-processed Applications object available throughout the app.
```

---

**Applications data structure**
- [x] model
- [x] schema
- [x] fixture
- [x] merged

scottnath:
Setting up Applications for Punchcard's database and processing through Content-Types.

**note:** _large_ PR due to fixtures

---
Resolves #430 

`DCO 1.1 Signed-off-by: Scott Nath<github@scottnath.com>`

### DIFF
--- a/lib/applications.js
+++ b/lib/applications.js
@@ -1,0 +1,7 @@
+'use strict';
+
+const model = require('./applications/model');
+
+module.exports = {
+  model,
+};
--- a/lib/applications.js
+++ b/lib/applications.js
@@ -1,7 +1,0 @@
-'use strict';
-
-const model = require('./applications/model');
-
-module.exports = {
-  model,
-};
--- a/lib/applications/model.js
+++ b/lib/applications/model.js
@@ -28,9 +28,9 @@ const structure = {
       description: 'Application name',
       inputs: {
         text: {
-          placeholder: 'Application name'
-        }
-      }
+          placeholder: 'Application name',
+        },
+      },
     },
     {
       type: 'text',
--- a/lib/applications/model.js
+++ b/lib/applications/model.js
@@ -33,34 +33,22 @@ const structure = {
       },
     },
     {
-      type: 'text',
+      type: 'url',
       id: 'live-endpoint',
       name: 'Live Endpoint',
       description: 'endpoint to call when content goes live',
     },
     {
-      type: 'text',
+      type: 'url',
       id: 'updated-endpoint',
       name: 'Updated Endpoint',
       description: 'endpoint to call when live content gets updated',
     },
     {
-      type: 'text',
+      type: 'url',
       id: 'sunset-endpoint',
       name: 'Sunset Endpoint',
       description: 'endpoint to call when live content gets sunsetted',
-    },
-    {
-      type: 'text',
-      id: 'client-id',
-      name: 'Client ID',
-      description: 'do not edit',
-    },
-    {
-      type: 'text',
-      id: 'client-secret',
-      name: 'Client Secret',
-      description: 'do not edit',
     },
   ],
 };
--- a/lib/applications/model.js
+++ b/lib/applications/model.js
@@ -1,0 +1,89 @@
+/**
+ *  @fileoverview Applications model
+ *
+ *  @author  Scott Nath
+ *
+ */
+'use strict';
+
+const types = require('punchcard-content-types');
+const config = require('config');
+
+/**
+ * Applications data entry structure
+ *
+ * @returns {object} - content structure to match requirements of punchcard-content-types
+ */
+const structure = {
+  name: 'Applications',
+  description: 'Contains webhook applications',
+  id: 'applications',
+  identifier: 'name',
+  attributes: [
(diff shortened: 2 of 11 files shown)


lib/applications/model.js
```diff
@@ -0,0 +1,89 @@
...
+      required: 'save',
+      description: 'Application name',
+      inputs: {
+        text: {
+          placeholder: 'Application name',
+        },
+      },
+    },
+    {
+      type: 'text',
```
Snugug: Endpoint should be `type: 'url'`

lib/applications/model.js
```diff
@@ -0,0 +1,89 @@
...
+      description: 'endpoint to call when live content gets updated',
+    },
+    {
+      type: 'text',
+      id: 'sunset-endpoint',
+      name: 'Sunset Endpoint',
+      description: 'endpoint to call when live content gets sunsetted',
+    },
+    {
+      type: 'text',
```
Snugug: Users don't get to edit this; why is it part of the form? This should really on't be displayed in the `view`.

lib/applications.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+const model = require('./applications/model');
+
+module.exports = {
+  model,
+};
```
Snugug: Can we start to follow the pattern in https://github.com/punchcard-cms/punchcard/issues/328 and have this be `lib/applications/index.js` instead of `lib/applications.js`?

lib/applications/model.js
scottnath: I see. The view is starting to get crowded. Perhaps it be displayed, as a non-editable item, in the edit screen instead. 
Regardless, we can discuss when the routing PR comes through.

lib/applications.js
scottnath: and remove `lib/applications.js`  - will do

scottnath:
@Snugug changes complete. Tests running

tests/applications.js
```diff
@@ -0,0 +1,54 @@
+import test from 'ava';
+import _ from 'lodash';
+
+import applications from '../lib/applications';
+import merged from './fixtures/applications/objects/model-merged.js';
+
+const reqObj = {
```
Snugug: Two things:
* Can this be renamed `req` as it doesn't conflict with anything else
* This is only used in one place, can it be moved to the one function it's used in?

Snugug:
@scottnath One last comment, then GTG

tests/applications.js
scottnath: @Snugug This is a leftover from me splitting out my massive, massive PR. I will be using this object extensively in the routes testing that is coming up. Within that testing `req` is commonly used.

Snugug: Sounds good

PR #439, opened 2016-09-20
Feature/local database

### LINKED ISSUE #283: Database for CI and local development
A large problem we run into is the use of postgres when creating our testing. When developing locally it writes to the local database and when errors happen, those entries are not automatically removed. For CI, the postgres timeouts often lead to failing builds - when there is nothing wrong with the build.

```gherkin
Feature: Data storage alternative for testing
  As a core developer
  I want data storage for development purposes
  So that I can develop my application with minimal storage issues.

  Scenario: Local development
    Given that we must develop locally
      And that we need to continually run tests while developing
      And that using the localhost database may cause errors when tests fail and do not trigger a removal of test data
      And that when running tests we need the option of adding and removing entries
    When I am running unit tests locally
    Then I need a data storage option that mimics the functionality of our production database
      And this option needs to seamlessly work with knex
      And this option should be configurable via the global configuration
      And this option should be separate from the database used when Punchcard is running on localhost.      
```

---

**Data storage alternative for testing**
- [ ] Local

scottnath:
I need to write to the local database, but only for unit testing purposes. I need this db to be different from the existing local database because it will be used throughout testing and should not cause un-intended errors when developing locally.
---
Resolves #283 

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ npm i punchcard-cms --save
 ```
 
 ```javascript
-const punchcard = require('punchcard');
+const punchcard = require('punchcard-cms');
 
 punchcard().then(app => {
   app.listen(8080);
@@ -35,15 +35,35 @@ $ npm run dev
 ## Setting up PostgreSQL locally
 
 1. Make sure you have homebrew installed
-`/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"`
+
+  `/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"`
+
 2. `brew update`
 3. `brew install postgres`
-4. type this to start the postgres server `postgres -D /usr/local/var/postgres`
+4. start the postgres server
+
+  `postgres -D /usr/local/var/postgres`
+
 5. open new terminal tab
-6.` createuser punchcard -P` (it'll prompt for a password) use `W@ts0n`
-7. `createdb -O punchcard punchcard`
-8. reconnect to psql as new user `psql -d punchcard -U punchcard`
-9. connect to db `\connect punchcard`
+6. create a postgres user (it'll prompt for a password) use `W@ts0n`
+
+  `createuser punchcard -P`
+
+7. Create local CMS database
+
+  `createdb -O punchcard punchcard`
+
+8. Create local test database
+
+  `createdb -O punchcard punchcard_test`
+
+9. reconnect to psql as new user
+
+  `psql -d punchcard -U punchcard`
+
+10. connect to db
+
+  `\connect punchcard`
 
 Once you are connected you should be able to use psql cli syntax to check tables.
 * `\d+` checks table relations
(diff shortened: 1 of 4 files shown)


Snugug:
**THAT IS A LOT OF BOLD CAPS**

README.md
```diff
@@ -35,15 +35,35 @@ $ npm run dev
 ## Setting up PostgreSQL locally
 
 1. Make sure you have homebrew installed
-`/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"`
+
```
Snugug: Not a fan of how these render; can we make this a little more in-line?

README.md
```diff
@@ -35,15 +35,35 @@ $ npm run dev
...
+  `/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"`
+
 2. `brew update`
 3. `brew install postgres`
-4. type this to start the postgres server `postgres -D /usr/local/var/postgres`
+4. start the postgres server
```
Snugug: Like here, `Start the Postgres server: postgres -D`...

tests/database.js
```diff
@@ -1,8 +1,19 @@
...
+import database from '../lib/database';
 
 test('Database Initialization', t => {
-  return db.init().then(() => {
+  return database.init().then(() => {
     t.pass();
+  }).catch(e => {
+    if (_.includes(e.message, `database "${config.knex.connection.database}" does not exist`)) {
+      // warning added for safety
+      console.log('--------------------------------------------------------------------------'); // eslint-disable-line no-console
```
Snugug: wow dat error message tho

scottnath:
WHEN I TYPE IT'S IMPORTANT. RECOGNIZE!

tests/database.js
scottnath: Can we keep it? Huh? Can we?
This need for a separate database is new. I just thought I'd gussy up the warning.

Changes desired? Or no warning at all?

Snugug: I mean, that's a little much.

Snugug: Also, `console.error` or `console.warn`

tests/database.js
```diff
@@ -12,6 +12,7 @@ test('Database Initialization', t => {
       // warning added for safety
       console.log('--------------------------------------------------------------------------'); // eslint-disable-line no-console
       console.log(`WARNING: you MUST create the ${config.knex.connection.database} database. See Punchcard README for details.`); // eslint-disable-line no-console
+      console.log(`OMG THE WORLD IS BURNING!!!`);
```
Snugug: burning

Ya know what? Isfine

tests/database.js
```diff
@@ -12,7 +12,7 @@ test('Database Initialization', t => {
       // warning added for safety
       console.log('--------------------------------------------------------------------------'); // eslint-disable-line no-console
       console.log(`WARNING: you MUST create the ${config.knex.connection.database} database. See Punchcard README for details.`); // eslint-disable-line no-console
-      console.log(`OMG THE WORLD IS BURNING!!!`);
+      console.log(`OMG THE WORLD IS BURNING!!!`); // eslint-disable-line no-console
```
Snugug: This needs to be single quotes instead of template strings because there are no templates in here

Issue #413, opened 2016-09-16
Send Webhook


scottnath:
When content gets sunrise and sunset, we should provide a way to reach out to an application and let them know they need to refresh their content

```gherkin
Feature: Send Webhook
  As an application developer
  I want to consume an event when content changes
  So that I can handle content changes automatically

  Scenario: Live
    Given that I have an endpoint set up for content going live
     When content is published in to the live API without currently being there
     Then when webhook endpoint should be called
      And it should contain its Client ID in the request body
      And it should contain its Client Secret in the request body
      And if the call isn't successful, the registering administrator should be notified

  Scenario: Update
    Given that I have an endpoint set up for content being updated
     When content in the live API is updated
     Then when webhook endpoint should be called
      And it should contain its Client ID in the request body
      And it should contain its Client Secret in the request body
      And if the call isn't successful, the registering administrator should be notified

  Scenario: Sunset
    Given that I have an endpoint set up for content being sunset
     When content in the live API is removed
     Then when webhook endpoint should be called
      And it should contain its Client ID in the request body
      And it should contain its Client Secret in the request body
      And if the call isn't successful, the registering administrator should be notified
```

---

**Send Webhook**
- [ ] Live
- [ ] Update
- [ ] Sunset

scottnath:
@Snugug I'd like to adjust these scenarios to:

```gherkin
...
     Then my webhook endpoint should be called
      And the call should contain its Client ID in the request body
      And the call should contain its Client Secret in the request body
      And if the call isn't successful, the response from my endpoint should be captured
      And the time of the response from my endpoint should be captured.
```

Issue #115, opened 2016-07-18
Refresh Webhook


Snugug:
When content gets sunrise and sunset, we should provide a way to reach out to an application and let them know they need to refresh their content

```gherkin
Feature: Register Application
  As an administrator
  I want to register endpoint to call when content has changed
  So that application developers can handle content changes automatically

  Scenario: Register
    Given that I know the URLs of endpoints to call on content actions
     When I navigate to /applications/new
     Then I can add a new application
      And I can set the application's name, which is required
      And I can set an endpoint to call when content goes live
      And I can set an endpoint to call when live content gets updated
      And I can set an endpoint to call when live content gets sunset 
      And I will receive a randomly generated, unique, hard to guess Client ID
      And I will receive a randomly generated, unique, hard to guess Client Secret
      And it will be assigned an internal ID
      And I will be redirected to its individual view page once saved

  Scenario: View All
    Given that I have one or more applications registered
     When I navigate to /applications
     Then I can see a list of all of the applications registered
      And I can see their names
      And I can see if they have a go live endpoint registered
      And I can see if they have an update endpoint registered
      And I can see if they have a sunset endpoint registered
      And I can see a black space if an endpoint is not registered
      And I can see an indicator if the the endpoint was successfully called or not if an endpoint was registered
      And the indicator will link to the edit view of the application showing history of that endpoint's calls
      And I will have a link to edit the application
      And the list of applications will be paginated in groups of 10

  Scenario: Add New
    Given that I wish to register an application
     When I navigate to /applications
     Then I click on a button that will allow me to add a new application
      And if there is a list of registered applications, the button will be above them

  Scenario: View One
    Given that I have an application registered
     When I navigate to /applications/:id
     Then I can see the application's name
      And I can see each of the application's endpoints
      And both the Client ID and Client Secret will be displayed
      And I can reset the Client Secret
      And I can update the application's name or endpoints
      And I can see the call history of each endpoint which includes the time it was run and if the call was successful or not
      And I can delete the application


```

---

**Register Webhook**
- [x] Register
- [x] View All
- [x] Add New
- [x] View One

Snugug:
Much of this is going to supersede https://github.com/punchcard-cms/punchcard/issues/119 but this is going to take precedence. https://github.com/punchcard-cms/punchcard/issues/119 should be turned in to allowing collaboration on applications.

This is super important for this sprint.

Snugug:
[...]

PR #442, opened 2016-09-21
Feature/apps templates

### LINKED ISSUE #441: Webhook templates
Creating templates for application management pages

```gherkin
Feature: App management templates
  As a core developer
  I want nunjucks templates for my applications management pages
  So that I can show users the content coming from the applications management section.
  
  Scenario: Common site elements
    Given that I am creating a nunjucks page template
      And that templates in Punchcard must contain common elements which represent the site's template
    When I am setting up a new page template in Punchcard
     Then I need the template to contain common site elements
      And that template must extend the site donut
      And that template must contain the page head
      And that template must contain the ux header
      And that template must contain the introduction head
      And that template must contain the footer

  Scenario: Landing
    Given that I am creating the landing page route
      And that route needs html
     When I am setting up Punchcard for the applications landing route
     Then I need a template for the applications landing page
      And the template must be written in nunjucks
      And the template must contain common site elements
      And the table must include a button to add a new application
      And it must include a table of all of the applications registered
      And the table must have a column for application name
      And the table must have a column for go live endpoint registered
      And the table must have a column for update endpoint registered
      And the table must have a column for sunset endpoint registered
      And the table must have a column with a link to edit the application.

  Scenario: Add New
    Given that I am creating the add new application route
      And that route needs html
    When I am setting up Punchcard for the applications new app route
     Then I need a template for the add new application page
      And the template must be written in nunjucks
      And the template must contain common site elements
      And the template must contain a form for adding a new app
      And said form should be a separate template
      And there should be javascript for the form in the footer created from Content-Types

  Scenario: Edit One
    Given that I am creating the application route to manage one application
      And that route needs html
    When I am setting up Punchcard for the manage app route
     Then I need a template for the manage application page
      And the template must be written in nunjucks
      And the template must contain common site elements
      And the title of the page should be `Edit`
      And the template must contain a form for editing an app
      And said form should be a separate template
      And there should be javascript for the form in the footer created from Content-Types

  Scenario: Form
    Given that I am creating the application routes to manage applications
      And those routes need html
      And management of applications requires a form
    When I am setting up applications edit or add pages
    Then I need a template for the form
      And the template should include a form
      And the form must have an action attribute determined by the routing
      And the form's inputs should be created by the Content-Types module
      And the form's actions must include at least one submit button
      And the form's actions must include at least one cancel button
      And the form's actions must include a submit button which says `Save` on the Add route
      And the form's actions must include a submit button which says `Update` on the One route
      And the form's actions must include a submit button which says `Delete` on the One route

```

---

**App management templates**
- [x] Common site elements
- [x] Landing
- [x] Add
- [x] One
- [x] Form

scottnath:
Application templates.

---
Resolves #441 


views/applications/_form.html
```diff
@@ -0,0 +1,13 @@
...
+  </div>
+
+  <div class="form--actions">
+    <input type="submit" name="submit" value="{{button}}" class="form--submit">
+    {% if delete %}
+      <input type="submit" name="submit" value="{{config.actions.delete}}" class="form--submit">
+    {% endif %}
+    <a href="../" class="form--cancel">Cancel</a>
+  </div>
+</form>
```
Snugug: I feel that this is likely shared a few places, like the User form? Can we consolidate down to one form?

Snugug: And form scripts should probably come along for the ride with the form

views/applications/all.html
```diff
@@ -0,0 +1,49 @@
...
+
+{% block header %}
+  {{ super() }}
+{% endblock %}
+
+{% block main %}
+  <div class="introduction">
+    <h1 class="introduction--h1">{{structure.name}}</h1>
+    <p class="introduction--p">{{structure.description}}</p>
+    <a href="/${config.base}/${config.actions.add}" class="introduction--button">{{config.actions.add|capitalize}} {{config.base}}</a>
```
Snugug: That's now how we spit out variables in Nunjucks

Snugug: I also have no way of knowing if these variables are correct without seeing the controller.

views/applications/all.html
```diff
@@ -0,0 +1,49 @@
...
+        <th class="base--th">Sunset</th>
+        <th class="base--th">Edit</th>
+      </tr>
+    </thead>
+    <tbody class="base--tbody">
+    {% for a in applications %}
+      <tr class="base--tr">
+        <td data-th="Application" class="base--td">{{a.name}}</td>
+        <td data-th="Created" class="base--td">{{a.created}}</td>
+        <td data-th="Live" class="base--td">{{a['live-endpoint']}}</td>
```
Snugug: * We aren't outputting the actual URLs here, right? We're outputting a representation of whether or not they A exist and B if they've passed or failed the most recent attempt. This also means that it should be blank if there isn't one available
* These should be `live`, `updated`, and `sunset`

views/applications/all.html
```diff
@@ -0,0 +1,49 @@
...
+        <th class="base--th">Updated</th>
+        <th class="base--th">Sunset</th>
+        <th class="base--th">Edit</th>
+      </tr>
+    </thead>
+    <tbody class="base--tbody">
+    {% for a in applications %}
+      <tr class="base--tr">
+        <td data-th="Application" class="base--td">{{a.name}}</td>
+        <td data-th="Created" class="base--td">{{a.created}}</td>
```
Snugug: Why are we showing created? If we do, it needs to be formatted, not the string raw from the DB

views/applications/all.html
Snugug: These should also link to the respective run views in the `one` view

views/applications/one.html
```diff
@@ -0,0 +1,26 @@
...
+  {{ super() }}
+{% endblock %}
+
+{% block main %}
+  <div class="introduction">
+    <h1 class="introduction--h1">{{config.actions.edit|capitalize}}</h1>
+    <p class="introduction--p">{{structure.description}}</p>
+  </div>
+  {% include "applications/_form.html" %}
+{% endblock %}
```
Snugug: View (and story) is missing the display of the Client ID and Client Secret, as well as buttons for revoking and resetting the tokens.

views/applications/one.html
```diff
@@ -0,0 +1,26 @@
...
+
+{% block footer %}
+  {{ super() }}
+
+  <script>
```
Snugug: This repeated block is why form scripts should come with forms

views/applications/all.html
scottnath: Yeah, those variables in there was a mistake. Will fix

As far as the controller...the only way I can do that is to have a _large_ PR that contains these templates as well. Happy to do it, but this split was mainly done to satisfy the "200 line max" requirement.

views/applications/all.html
scottnath: ## urls
from #115:
```
      And I can see if they have a go live endpoint registered
      And I can see if they have an update endpoint registered
      And I can see if they have a sunset endpoint registered
```
nothing about pass/fail of recent attempt. These would be blank if empty, and the URL if not empty. Would you like it to rewrite the scenario to clarify? Or leave this as is and write a new scenario to add these features once the base version of this is complete?

## variables names
follows the variables set out [in the merged schema](https://github.com/punchcard-cms/punchcard/blob/master/lib/database/schemas.js#L137). I can change these, but would you like a schema change too?

## link
Only instruction in #115 is `And I will have a link to edit the application`. Is there a different story I'm missing for this?

## run views
What's a run view? Scenario for viewing and editing a single application:

```gherkin
  Scenario: View One
    Given that I have an application registered
     When I navigate to /applications/:id
     Then I can see the application's name
      And I can see each of the application's endpoints
      And both the Client ID and Client Secret will be displayed
      And I can reset the Client Secret
      And I can update the application's name or endpoints
      And I can delete the application
```

views/applications/_form.html
scottnath: Consolidation is out of scope, but I can adjust this to be ready for that consolidation.

views/applications/all.html
scottnath: will remove

views/applications/all.html
scottnath: variables fixed

views/applications/all.html
scottnath: created removed

views/applications/one.html
scottnath: Left out of story on purpose. Two pieces - pagination and these revokations are my last pieces of the puzzle and still tbdone.

views/applications/_form.html
Snugug: Well my point is [that already exist](https://github.com/punchcard-cms/punchcard/blob/master/views/users/_form.html) and is identical, save some spacing and the `delete` if statement. Instead of duplicating work, update that one, and use it. We can spin them out into a shared basic form later

views/applications/all.html
Snugug: * URL: We had discussed this in person, but it appears as if it wasn't captured. Displaying the URLs here would be all but meaningless to our users. I'll update the scenario to reflect the discussion. I've updated the scenarios to reflect the conversation we had
* The schema can stay the same, but we should use something that's more human-usable in our templates
* I've updated #115 to reflect the conversation we had about audit trail of webhook fire history.

views/applications/one.html
Snugug: I'm OK leaving pagination out of this PR until we figure out a more universal solution for for paginating the tables we have, but the Client ID and Client Secret are hard requirements for view; they're literally the thing we need to get from the webhooks from a user perspective.

views/applications/_form.html
scottnath: @Snugug but delete _is_ different for this form _and_ [users has a different way to currently handle deletes](https://github.com/punchcard-cms/punchcard/blob/master/lib/routes/users.js#L189)
The stories I'm writing for the webhooks will be copied when updating the Users section - but we should absolutely not attempt to make changes to the users section in this pr - that is very Out Of Scope.

views/_form.html
```diff
@@ -0,0 +1,18 @@
+<form action="{{action}}" method="post" enctype="multipart/form-data" class="form" novalidate>
```
Snugug: We can start arranging our work a little better. Can this be put in `views/partials` or `views/components` and rename it something like `_basic-form.html`

views/applications/all.html
scottnath: @Snugug your changes to the story would very much up the estimate I would make on that story _and_ would greatly update the size of this PR. Adding in the responses from webhooks or using those responses in this view page should be separate, possibly part of #413 or possibly their own story alltogether. I request you revert the story. At this point, you're moving my goal posts while I'm trying to run down the field

scottnath: template vars: What would you like them to be? For this developer, `live-endpoint` tells me a lot more than `live`.

Snugug: `a.endpoints.live`. Make it an object. Also, any reason why you use `a` instead of `app`? Single letter variables we tend to want to avoid

views/_form.html
scottnath: I'll use `_basic-form.html`

views/applications/all.html
Snugug: @scottnath I'm not moving the goalpost, this is the conversation we had Wednesday and Thursday about this view while you were looking for advice for implementing. For _just_ the views, this doesn't increase scope in a significant way. This is _just_ the views. It may wind up increasing the scope of the the controller story, but not by much (the audit trail already in the DB will provide that information), grabbing the latest entry and seeing if it passed or not, and then passing to the view if it indeed passed or failed with an if statement accordingly

views/applications/all.html
scottnath: `here` in the template or `here` in this PR's comments?

As for this particular line, it follows the [structure already established for content](https://github.com/punchcard-cms/punchcard/blob/master/views/content/landing.html#L14)

views/applications/all.html
Snugug: `{% if a.endpoints.live %} {{a.endpoints.live}} {% endif %}` where `a.endpoints.live` is either `'pass'`, `'fail'` based on the results of the webhook (from the audit trail). Wrap it in a span with a `pass/fail` class so we have styling hooks later on. That's the entire extent of this addition. Or, if you'd like, `✓` for pass and `✗` for fail

Snugug: It appears as if I left out the links for this. My bad. 
`{% if a.endpoints.live %}<a href="/{{config.base}}/{{a.id}}#live" class="base--link"><span class="endpoint-{{a.endpoints.live}}">{{a.endpoints.live}}</span></a>`

Snugug: Then, `applications/one.html`, after the form, loop over each endpoint, build a table with `success` and `timestamp` headers, with those spit out (maybe only the last 5? 10?) and give them an anchor.

Like I said, for a `views` story, this does not add much

scottnath: ## client changes
So, I am not in any way saying that I am not doing that. I am saying that I am trying to not do that _in this PR_. I will absolutely be building out that client id/secret functionality. This PR, which has taken up great amounts of my time today, was _only_ created to satisfy the 200 line limit for our PRs. If I did not have that limit, I would not have done any of this. And you would also have had the routes to look at to determine the variables a template would receive. js

And yes, sorry, but the goal post is moving. We did discuss the audit trail for this, but we also discussed it being a separate story.

scottnath: And for the code you've laid out above, I totally understand that the endpoints will pass or fail. but for right now, none of that functionality exists and I am just desperately trying to get the management of these applications into Punchcard - which does not include any of the actual functionality for running the webhooks - which is what would give the pass/fail int he first place.

Snugug: Which means that including this now will have it available in our view when that functionality is added.

scottnath: Please stop looking for the final solution in this PR. I wrote the stories on purpose and split #115 on purpose. Management of a new type of data in punchcard requires a lot of moving parts which are somewhat generic in nature. Schemas/models/templates/forms. All these are required to be put together before any of the meat of this system can be dug into.

scottnath:
@Snugug changes added

views/applications/one.html
```diff
@@ -0,0 +1,102 @@
...
+{% endblock %}
+
+{% block main %}
+  <div class="introduction">
+    <h1 class="introduction--h1">{{config.actions.edit|capitalize}}</h1>
+    <p class="introduction--p">{{structure.description}}</p>
+  </div>
+  {% include "_basic-form.html" %}
+
+  {% if app['client-id'] %}
```
Snugug: Would like to see this be similar to the endpoints from above; either `app.client` and `app.secret` or `app.client.id` and `app.client.secret`

views/applications/one.html
```diff
@@ -0,0 +1,102 @@
...
+    <dl class="base--dl">
+      <dt class="base--dt">Client ID</dt>
+      <dd class="base--dd">app['client-id']</dd>
+      <dd class="base--dd"><a href="#TBD">Create new Client ID</a></dd>
```
Snugug: Per #115, can't regen Client ID, only secret.

views/applications/one.html
```diff
@@ -0,0 +1,102 @@
...
+    </dl>
+  {% endif %}
+
+  {% if app['client-secret'] %}
+    <dl class="base--dl">
+      <dt class="base--dt">Client Secret</dt>
+      <dd class="base--dd">app['client-secret']</dd>
+      <dd class="base--dd"><a href="#TBD">Create new Client Secret</a></dd>
```
Snugug: Link to create new secret isn't part of the definition of a client secret

views/applications/all.html
```diff
@@ -0,0 +1,47 @@
...
+        <th class="base--th">Updated</th>
+        <th class="base--th">Sunset</th>
+        <th class="base--th">Edit</th>
+      </tr>
+    </thead>
+    <tbody class="base--tbody">
+    {% for app in applications %}
+      <tr class="base--tr">
+        <td data-th="Application" class="base--td">{{app.name}}</td>
+        <td data-th="Live" class="base--td">{% if app.endpoints.live %}<a href="/{{config.base}}/{{a.id}}#live" class="base--link"><span class="endpoint-{{app.endpoints.live|first}}">{{app.endpoints.live|first}}</span></a>{% endif %}</td>
```
Snugug: Are you expecting to have the same data structure here as you are in `one.html`? If so, these need to be grabbing `response` property of the first item, not just the first item.

views/applications/all.html
```diff
@@ -30,9 +30,9 @@ <h1 class="introduction--h1">{{structure.name}}</h1>
     {% for app in applications %}
       <tr class="base--tr">
         <td data-th="Application" class="base--td">{{app.name}}</td>
-        <td data-th="Live" class="base--td">{% if app.endpoints.live %}<a href="/{{config.base}}/{{a.id}}#live" class="base--link"><span class="endpoint-{{app.endpoints.live|first}}">{{app.endpoints.live|first}}</span></a>{% endif %}</td>
-        <td data-th="Updated" class="base--td">{% if app.endpoints.updated %}<a href="/{{config.base}}/{{a.id}}#updated" class="base--link"><span class="endpoint-{{app.endpoints.updated|first}}">{{app.endpoints.updated|first}}</span></a>{% endif %}</td>
-        <td data-th="Sunset" class="base--td">{% if app.endpoints.sunset %}<a href="/{{config.base}}/{{a.id}}#sunset" class="base--link"><span class="endpoint-{{app.endpoints.sunset|first}}">{{app.endpoints.sunset|first}}</span></a>{% endif %}</td>
+        <td data-th="Live" class="base--td">{% if app.endpoints.live %}<a href="/{{config.base}}/{{a.id}}#live" class="base--link"><span class="endpoint-{{app.endpoints.live.response|first}}">{{app.endpoints.live.response|first}}</span></a>{% endif %}</td>
```
Snugug: I'm not sure if `app.endpoints.live.response|first` will actually get the first item in `app.endpoints.live` and then the `response` property, but we can deal with that when we've got actual data in there if it doesn't work

PR #443, opened 2016-09-21
Fix select reference not having a placeholder

### LINKED ISSUE #381: Bug: Reference plugin select input has no default option
Reference plugin select input has no default option. It just selects the first item in the list. 

## Steps for Reproducing

- Add/Edit a revision of content with a reference (Product)
- Go to the reference (Taxonomy - Status)
  - Observe that the first option is `General Availability (GA)` and not `-Select a Status-`.

## Screenshots

image

## Affected Browsers

- [ ] All

## Runtime Version

5.11

## Code Version

2.10.0

Snugug:
Adds default select option for reference

---
Resolves #381 

### DIFF
--- a/input-plugins/input-plugin-reference/index.js
+++ b/input-plugins/input-plugin-reference/index.js
@@ -30,7 +30,6 @@ module.exports = {
       label: 'Select from the following',
       options: [],
       settings: {
-        empty: false,
         view: 'select',
         placeholder: '-- Choose One --',
       },
--- a/input-plugins/input-plugin-reference/index.js
+++ b/input-plugins/input-plugin-reference/index.js
@@ -31,7 +31,8 @@ module.exports = {
       options: [],
       settings: {
         empty: false,
-        view: 'select',    
+        view: 'select',
+        placeholder: '-- Choose One --',
       },
     },
   },
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -327,10 +327,6 @@ const references = (types) => {
  */
 const fill = (types, type, refs, database) => {
   const ct = _.cloneDeep(type);
-  const placeholder = {
-    value: '',
-    label: '-- Choose One --',
-  };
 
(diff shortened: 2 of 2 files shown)


lib/utils.js
```diff
@@ -327,6 +327,11 @@ const references = (types) => {
  */
 const fill = (types, type, refs, database) => {
   const ct = _.cloneDeep(type);
+  const placeholder = {
+    value: '',
+    label: '-- Choose One --',
```
scottnath: There should be an optional way to change this text. Can you please add a [setting option for references](https://github.com/punchcard-cms/punchcard/blob/master/input-plugins/input-plugin-reference/index.js#L32) with this text as default?

Snugug: Yup!

Snugug: @scottnath Take a look, updated

input-plugins/input-plugin-reference/index.js
```diff
@@ -31,9 +31,10 @@ module.exports = {
       options: [],
       settings: {
         empty: false,
```
scottnath: can you delete this line while in here please?

scottnath:
@Snugug small thing, then GTG!

input-plugins/input-plugin-reference/index.js
Snugug: THAT'S OUT OF SCOPE!
BUT THAT'S OK

Snugug: While I'm in there, should I write validation to ensure that the item chosen is one of the options selected? Or should we write an actual story for that?

scottnath: I think a story for that please! I think there is more expansion on this plugin than just that, so let that part stay OOScope for now...but not my `remove empty` request...that's the good kind of OOScope - ie very smartypants.

PR #448, opened 2016-09-23
apps utils and tests

scottnath:
@Snugug @joshblack 
This is primarily a break out of the tests which come in at 459 lines for the applications section. A couple new other functions included, as well as the scaffolding for the routing.

I have the routing completely finished as well, but can't submit until this has been gone over. 

Related: #115 

---


Snugug:
@scottnath What issue is this in relation to? Even if it doesn't close the issue, please link to it

tests/applications.js
```diff
@@ -1,28 +1,134 @@
...
+};
+
+const date = moment().format('YYYY-MM-DD hh:mm');
+
+const body = {
+  'name--text': 'Bar',
+  'live-endpoint--url': 'http://bar.com/live',
+  'updated-endpoint--url': 'http://bar.com/updated',
+  'sunset-endpoint--url': 'http://bar.com/sunset',
+  'client-id': uuid.v4(),
```
Snugug: These aren't fields; they're [generated and just displayed](https://github.com/punchcard-cms/punchcard/blob/master/views/applications/all.html#L19-L40). Should not be included in the test body.

tests/applications.js
```diff
@@ -69,3 +182,352 @@ test('Workflow model from config', t => {
...
+  req.body = _.cloneDeep(body);
+
+  const request = httpMocks.createRequest(req);
+
+  const response = httpMocks.createResponse({ eventEmitter: EventEmitter });
+  applications.routes.save(request, response);
+  response.render();
+
+  response.on('end', () => {
+    t.is(response.statusCode, 302, 'Should be a 302 response');
```
Snugug: I'd like to see this redirected to the application individual page as per our scenarios

Snugug: I'd also like to see a test to ensure that it has been assigned an internal ID (which I guess would go hand-in-hand with ensuring the redirection works)

tests/applications.js
```diff
@@ -69,3 +182,352 @@ test('Workflow model from config', t => {
...
+    t.is(response._getRedirectUrl(), '/applications');
+    t.end();
+  });
+});
+
```
Snugug: There seems to be missing a test for reseting Client Secret

lib/applications/utils.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
+
+/**
+ * @fileoverview Applications utility functions
+ *
+ */
+const _ = require('lodash');
+
+const endpoints = rw => {
+  const row = rw || {};
```
scottnath: This entire file is now removed.

tests/applications.js
scottnath: removed

tests/applications.js
scottnath: see above `Routes - Secret`

tests/applications.js
scottnath: added a check of the redirect url to make sure that it is redirecting to `/applications/some-number` by checking the parts of the url. The only way to know the exact number though would be to do a database call.

scottnath:
@Snugug all comments addressed. Removed _entire_ utils file -- wasn't needed.

scottnath:
@Snugug big massive to do for mocking databases:
https://github.com/punchcard-cms/punchcard/issues/455

Issue #218, opened 2016-08-09
Core Dev: Content routing E2E refactor


scottnath:
The test file, server.js, in punchcard tests is causing a great deal of issues. We need to remove all testing which does anything more than check the existence and viability of the routes.

```gherkin
Feature: Content routing unit testing
  As a core developer
  I want to create unit tests for all non-routing code in the content routing
  So that our code is confirmed by direct unit testing

  Scenario: Discover uncovered lines
    Given that the routing is being tested as E2E instead of the code being unit testing
     When I remove the /content routes from server.js
     Then I should have a clear indication of which lines need to be covered via unit testing

  Scenario: Refactor uncovered lines
    Given that some code can be broken out of routes
     When I move code to other files
     Then I should create new tests to cover that moved code.
```

---

**Unit Testing**
- [ ] Discovery
- [ ] Refactor as needed

Snugug:
This is becoming a large issue for us as these tests fail intermittently and it negatively affects our ability reliably release deploys.

scottnath:
Continued refactor is blocked by the need for a testing database option:
https://github.com/punchcard-cms/punchcard/issues/283

scottnath:
note-to-self: research turning `utils.content.check.url` into middleware. Middleware would need to change express.req

scottnath:
note-to-self II: refactor should test that the `action` attribute on the response object is the what we expect. This will be done with `node-mocks-http`.

PR #450, opened 2016-09-26
Hotfix/add save bug

### LINKED ISSUE #447: Bug: Routing: there is a too-open search for `add` to change to `save`
If a content type contains the word `add`, the routing attempts to replace it with the word `save`

## Steps for Reproducing

- Create a content type with the word `add` in the name, like `add-on`
- Attempt to save an `add-on` in the CMS
- Note that it results in a 404 page because it can't find the content type `save-on`

## Screenshots
image

## Affected Browsers

- [ ] All

## Runtime Version

5.11

## Code Version

2.10.0
---

scottnath:
Error was with how the `action` was created for the form on the `add` page. Confirmed repair with integration test.

---
Resolves #447 


lib/routes/content.js
```diff
@@ -215,7 +215,7 @@ const routes = application => {
             return types.form(merged, errors, config).then(form => {
               res.render('content/add', {
                 form,
-                action: `/${config.content.base}/${req.params.type}/${config.content.actions.save}`,
+                action: `/${config.content.base}/${req.params.type.toLowerCase()}/${config.content.actions.save}`,
```
Snugug: This all looks good. Just requesting that a few tests are written so we can catch this before it crops up again?

scottnath: @Snugug cannot test this because the routes are not yet pulled out of the `content` section. [this comment on the refactor story](https://github.com/punchcard-cms/punchcard/issues/218#issuecomment-249609951) captures the need to add a test that the `action` attribute is what we expect it to be.

ie...cannot add a test now, unless we make it an E2E test in `server.js`. Which is dumb. I hate that file. I haet
ihaetusohard
 that file so hard.

PR #456, opened 2016-09-26
Feature/webhook admin

### LINKED ISSUE #115: Refresh Webhook

scottnath:
@Snugug @joshblack 

This is the final PR for creating webhooks management in Punchcard. It sure is big! Good thing my tests didn't need changes except pulling out the `skip`s ;-)

---
Resolves #115 


lib/applications/routes.js
```diff
@@ -4,53 +4,322 @@
...
  * @param {object} res - HTTP Response
  * @param {object} next - Express callback
  *
+
+ * @returns {promise} promise wrapping database promise
  */
-const all = () => {};
+const all = (req, res, next) => {
+  return new Promise((resolve, reject) => {
+    return database
```
Snugug: Database is a promise, don't wrap it in a promise

lib/applications/routes.js
```diff
@@ -4,53 +4,322 @@
...
+      .select('*')
+      .from(`${config.applications.base}`)
```
Snugug: Our Database table names aren't configurable. This is [always going to be `applications`](https://github.com/punchcard-cms/punchcard/blob/master/lib/database/schemas.js#L238-L239). This is an incorrect use of configuration

lib/applications/routes.js
```diff
@@ -4,53 +4,322 @@
...
+      .orderBy('name', 'DESC').then(rws => {
+        // add itentifier to each row
+        const rows = utils.routes.identifier(rws, req.applications.merged);
```
Snugug: This is _not_ our content. Why are you running this through the route identifier instead of just using the `name` column readily available in the rows you're going to be retrieving?

lib/applications/routes.js
```diff
@@ -4,53 +4,322 @@
...
  * https://github.com/punchcard-cms/content-types/issues/112
  */
-const add = () => {};
+const add = (req, res, next) => {
+  const errors = _.get(req.session, 'form.applications.save.errors', {});
+  const values = _.get(req.session, 'form.applications.save.content', {});
+
+  _.unset(req.session, 'form.applications.save');
+
+  return types.only(config.applications.base, values, [req.applications.merged], config).then(merged => {
```
Snugug: The `base` config _is not_ a reliable source for us to use for things like database tabes and content configuration. While that _may_ hold true, now, those need to be hard-coded values because if that changes in config, everything will break.

lib/applications/routes.js
```diff
@@ -4,53 +4,322 @@
...
+          return row;
+        });
+
+        res.render('applications/all', {
+          applications,
+          config: config.applications,
+          structure: model.structure,
+        });
+
+        resolve(true);
```
Snugug: Should `return true` for a success

lib/applications/routes.js
```diff
@@ -4,53 +4,322 @@
...
+      }).catch(e => {
+        const err = {
+          message: e.message,
+          safe: '/',
+          status: 500,
+        };
+
+        reject(err);
```
Snugug: Should `return next(err)` instead of rejecting

lib/applications/routes.js
```diff
@@ -4,53 +4,322 @@
...
  * View/edit an individual Webhook
  * @param {object} req - HTTP Request
  * @param {object} res - HTTP Response
  * @param {object} next - Express callback
  *
+ * @returns {promise} resolves/rejects promise
  */
-const one = () => {};
+const one = (req, res, next) => {
+  const id = _.get(req.session, 'form.applications.edit.id', null);
```
Snugug: Won't `id` come from `req.params.id`?

lib/applications/routes.js
```diff
@@ -4,53 +4,322 @@
...
+      safe: `/${config.applications.base}`,
+      status: 500,
+    };
+
+    return next(err);
+  });
+};
 
 /**
  * View/edit an individual Webhook
```
Snugug: Individual Application, not Webhook

lib/applications/routes.js
```diff
@@ -4,53 +4,322 @@
...
+  return new Promise((resolve, reject) => {
+    // something went wrong on save:
+    if (Object.keys(values).length > 0 && id) {
+      // add the previous session data back in
+      _.set(req.session, 'form.applications.edit', {
+        id,
+      });
+
+      // grab the edit form and inject errors and values from req.session
+      return types.only(config.applications.base, values, [req.applications.merged], config).then(merged => {
```
Snugug: Same `base` issue here

lib/applications/routes.js
```diff
@@ -4,53 +4,322 @@
...
+const one = (req, res, next) => {
+  const id = _.get(req.session, 'form.applications.edit.id', null);
+  const errors = _.get(req.session, 'form.applications.save.errors', {});
+  const values = _.get(req.session, 'form.applications.save.content', {});
+  let app = _.get(req.session, 'form.applications.edit.app', null);
+
+  _.unset(req.session, 'form.applications.edit');
+  _.unset(req.session, 'form.applications.save');
+
+  return new Promise((resolve, reject) => {
```
Snugug: Same promise wrap here as above

lib/applications/routes.js
```diff
@@ -4,53 +4,322 @@
...
-const secret = () => {};
+const secret = (req, res, next) => {
+  const id = _.get(req.session, 'form.applications.edit.id', null);
+  const referrer = req.get('Referrer');
+
+  const data = {
+    'client-secret': uuid.v4(),
+  };
+
+  return new Promise((resolve, reject) => {
```
Snugug: Promise wrap

lib/applications/routes.js
```diff
@@ -4,53 +4,322 @@
...
+      const err = {
+        message: 'Secret can only be changed from the application edit screen',
+        safe: `/${config.applications.base}`,
+        status: 500,
+      };
+
+      reject(err);
+    }
+
+    return database(`${config.applications.base}`).where('id', '=', id).update(data).then(() => {
```
Snugug: Same database issue

lib/applications/routes.js
```diff
@@ -4,53 +4,322 @@
...
+
+      reject(err);
+    });
+  }).catch(err => {
+    return next(err);
+  });
+};
 
 /**
  * View/edit an individual Webhook
```
Snugug: This should be `Save an application`

lib/applications/routes.js
```diff
@@ -4,53 +4,322 @@
...
+ * @returns {promise} database promise
  */
-const save = () => {};
+const save = (req, res) => {
+  const id = _.get(req.session, 'form.applications.edit.id', null);
+  const referrer = (_.get(req.session, 'referrer') || req.get('Referrer')) || `/${config.applications.base}`;
+
+  // user hit delete button
+  if (req.body.submit === config.applications.actions.delete && id) {
+    return database(`${config.applications.base}`).where('id', '=', id).del().then(() => {
```
Snugug: Config database name

lib/applications/routes.js
```diff
@@ -4,53 +4,322 @@
...
+
+        return res.redirect(`/${config.applications.base}`);
+      });
+    }
+
+    // on add, create fresh id/secret
+    data['client-id'] = uuid.v4();
+    data['client-secret'] = uuid.v4();
+
+    return database(`${config.applications.base}`).insert(data).returning('id').then((newid) => {
```
Snugug: Config database name

views/applications/all.html
```diff
@@ -12,7 +12,7 @@
   <div class="introduction">
     <h1 class="introduction--h1">{{structure.name}}</h1>
     <p class="introduction--p">{{structure.description}}</p>
-    <a href="/{{config.base}}/{{config.actions.add}}" class="introduction--button">{{config.actions.add|capitalize}} {{config.base}}</a>
+    <a href="/{{config.base}}/{{config.actions.add}}" class="introduction--button">{{config.actions.add|capitalize}} Application</a>
```
Snugug: Now why isn't `config.base` being used here?

views/applications/one.html
```diff
@@ -25,15 +26,16 @@ <h1 class="introduction--h1">{{config.actions.edit|capitalize}}</h1>
...
   {% endif %}
 
   {% if app.endpoints %}
     <h2>Endpoints</h2>
 
-    {% if app.endpoints.live %}
-      <h3 id="live">Live</h3>
+    {% if app.endpoints.live | length %}
+      <h3>Live</h3>
+      <a name="live"></a>
```
Snugug: Why were these changed from `id` to `name`? `id` is semantically correct and `name` [has been deprecated](https://www.w3.org/TR/html5/obsolete.html) on anchor tags. Please return this to having `id` on the `h3`

Snugug:
@scottnath Test are stalling at `applications › Single application route`

lib/applications/routes.js
scottnath: no, this id is specifically coming from session (or null). It is set in `save` and used here if save rejects back to `one`.

scottnath:
@Snugug so, I've finally got all my tests except one to pass, which I skipped. Literally the only diff with the one before it (that passes) and this one ('Save new application') is that knex does an insert and return. I think this test should be put on hold until we can come up with a mock of the database for testing.

Unfortunately, now tests for `schedule` are failing every time. It's either the `schedule > push` or `schedule > pull` test (not both.) One of them fails every time. Can you fix that please so I can complete this PR? Or should I skip them?

Snugug:
Skip them for now, they were passing (otherwise they wouldn't have found their way in) and we shouldn't block this PR on being able to track that down

lib/applications/routes.js
```diff
@@ -4,53 +4,317 @@
...
+  const errors = _.get(req.session, 'form.applications.save.errors', {});
+  const values = _.get(req.session, 'form.applications.save.content', {});
+
+  _.unset(req.session, 'form.applications.save');
+
+  return types.only('applications', values, [req.applications.merged], config).then(merged => {
+    return types.form(merged, errors, config).then(form => {
+      res.render('applications/add', {
+        form,
+        action: req.url.replace(config.applications.actions.add, config.applications.actions.save),
```
Snugug: This is what got us in to trouble with the content routes. Can we take the same tactic we did there for this action?

Snugug: And you have below

lib/applications/routes.js
Snugug: OK

views/applications/_form.html
```diff
@@ -0,0 +1,13 @@
+<form action="{{action}}" method="post" enctype="multipart/form-data" class="form" novalidate>
```
Snugug: Why this over the previously created [basic form](https://github.com/punchcard-cms/punchcard/blob/master/views/_basic-form.html)

tests/applications.js
```diff
@@ -117,7 +117,7 @@ test('Workflow model from config', t => {
 //////////////////////////////
 // Routes - Applications landing
 //////////////////////////////
-test.cb.skip('All applications route', t => {
+test.cb.serial('All applications route', t => {
```
scottnath: I'll remove the serial. that was from testing to find the issues

views/applications/_form.html
scottnath: should have been deleted. removing now.

PR #457, opened 2016-09-26
Add key and key_slug filters

### LINKED ISSUE #407: API Filtering
When accessing our APIs, a user may want to filter results based on things they know without knowing the specific item they're looking for

```gherkin
Feature: Filter APIs
  As an application developer
  I want to filter my API results
  So that I can find items in the API without knowing exact attributes

  Scenario: Key
    Given that I want to retrieve API items with a specific key
     When I call an API endpoint that gives me a listing of items
     Then I should be able to indicate that I only want items returned if its key matches what I have specified
      And I should be able to indicate that I only want items returned if its key_slug matches what I have specified
```

---

**Filter API**
- [ ] Key

Snugug:
{{short description}}

---
Resolves #407 

### DIFF
--- a/lib/api/apis.js
+++ b/lib/api/apis.js
@@ -63,7 +63,22 @@ const types = app => {
 const all = query => {
   const organize = utils.organize(query);
 
-  return database('live').select('*').orderBy(organize.sort.by, organize.sort.dir).offset(organize.page.offset).limit(organize.page.limit).then(rows => {
+  console.log(query);
+
+  const results = database('live').select('*');
+
+  if (query.key) {
+    results.where({
+      key: query.key,
+    });
+  }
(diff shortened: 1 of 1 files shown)


lib/api/apis.js
```diff
@@ -63,7 +63,22 @@ const types = app => {
 const all = query => {
   const organize = utils.organize(query);
 
-  return database('live').select('*').orderBy(organize.sort.by, organize.sort.dir).offset(organize.page.offset).limit(organize.page.limit).then(rows => {
+  console.log(query);
```
scottnath: @Snugug *dude*

Snugug: Already fixed

PR #469, opened 2016-09-30
Feature/webhook send

### LINKED ISSUE #413: Send Webhook

scottnath:
Adding application endpoint calling via request & schedule.

**note**: sunrise/sunset tests skipped cause they are brittle

---
Resolves #413


scottnath:
@Snugug FINALLY!

lib/applications/send.js
```diff
@@ -1,5 +1,6 @@
 'use strict';
 
+const Promise = require('bluebird');
```
Snugug: Why are you bringing Bluebird in here?

lib/applications/send.js
```diff
@@ -142,6 +143,12 @@ const requests = options => {
 const save = options => {
   const ids = options.endpoints.map(end => {
     return end.id;
+  }).filter(id => {
+    if (typeof id !== 'undefined') {
```
Snugug: Can we assert true things instead of false things?

lib/init/schedule.js
```diff
@@ -3,7 +3,7 @@
 const schedule = require('../schedule');
 
 module.exports = app => {
-  return schedule.bootstrap().then(() => {
+  return schedule.bootstrap(app.get('applications-apps')).then(() => {
```
Snugug: Did I really let `applications-apps` through?

lib/schedule/utils.js
```diff
@@ -69,7 +77,9 @@ const pull = revision => {
  *
  * @return Promise
  */
-const push = revision => {
+const push = (revision, apps) => {
+  let options;
```
Snugug: This can be a `const` with the `updated` options, then line 106 you can change the trigger just as-is. `const` (and our linting) doesn't about reassigning parameters, only the ident itself

PR #471, opened 2016-10-03
apps in memory; fixed save new app test

scottnath:
A break up of my larger PR.

This PR adds `applications` to the express app memory, and also adds it to the request object. System changes the object on applications change

---

### DIFF
--- a/lib/applications/routes.js
+++ b/lib/applications/routes.js
@@ -291,13 +291,22 @@ const save = (req, res) => {
     data['client-id'] = uuid.v4();
     data['client-secret'] = uuid.v4();
 
+    console.log('Test app reached past update and delete');
+    console.log(JSON.stringify(data, null, 2));
+
     return database('applications').insert(data).returning('*').then((added) => {
+    console.log('Test app reached into insert');
+    console.log(JSON.stringify(added, null, 2));
       _.unset(req.session, 'form.applications.save');
 
       // add this new app to the request's apps object
       req.applications.apps.push(added[0]);
+    console.log('Test app reached past req push');
+    console.log(JSON.stringify(req.applications.apps, null, 2));
 
       res.redirect(`/${config.applications.base}/${added[0].id}`);
+    console.log('Test app reached past res.redirect');
+    console.log(`/${config.applications.base}/${added[0].id}`);
 
       return true;
     });
--- a/lib/applications/routes.js
+++ b/lib/applications/routes.js
@@ -291,27 +291,15 @@ const save = (req, res) => {
     data['client-id'] = uuid.v4();
     data['client-secret'] = uuid.v4();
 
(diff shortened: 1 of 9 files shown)


lib/init/settings.js
```diff
@@ -24,6 +24,20 @@ module.exports = (app) => {
   }).then(merged => {
     app.set('applications-merged', merged[0]);
 
+    return database('applications').select('*').orderBy('name', 'DESC').then(rows => {
```
Snugug: I fell in to this trap with APIs. We can't grab these only during initialization, we need to grab these when they're needed. Otherwise, if they change _after_ initialization, we won't have the updated list.

lib/init.js
```diff
@@ -22,14 +22,14 @@ module.exports = (application) => {
...
   }).then(app => {
-    return schedule(app);
-  }).then(app => {
     return passport(app);
   }).then(app => {
     return acl(app);
   }).then(app => {
     return settings(app);
   }).then(app => {
+    return schedule(app);
```
Snugug: Why did you move schedule?

lib/applications/routes.js
```diff
@@ -299,10 +291,13 @@ const save = (req, res) => {
     data['client-id'] = uuid.v4();
     data['client-secret'] = uuid.v4();
 
-    return database('applications').insert(data).returning('id').then((newid) => {
+    return database('applications').insert(data).returning('*').then((added) => {
       _.unset(req.session, 'form.applications.save');
 
-      res.redirect(`/${config.applications.base}/${newid}`);
+      // add this new app to the request's apps object
+      req.applications.apps.push(added[0]);
```
Snugug: Does this actually update `app.get('applications')`?

lib/applications/routes.js
```diff
@@ -285,10 +269,18 @@ const save = (req, res) => {
...
         _.unset(req.session, 'form.applications.edit');
         _.unset(req.session, 'form.applications.save');
 
+        // find this app in the request apps object
+        const index = req.applications.apps.findIndex(app => {
+          return app.id === id;
+        });
+
+        // we WANT to update this app in the request's apps object
+        req.applications.apps[index] = updated[0]; // eslint-disable-line no-param-reassign
```
Snugug: Can we use lodash `_.set` instead of disabling our linter?

Snugug:
Discussed, going to convert from `app.get` and direct `req` manipulation in routes to middleware for the API routes

lib/init.js
Snugug: 🐗 

(no reason, just a fancy emoji I hadn't seen before)

lib/init/settings.js
scottnath: I've changed this now and I'm updating the express settings on change of apps.

PR #473, opened 2016-10-05
applications send

scottnath:
Another send-webhooks split. This one breaks out the `send` functions.

relates to #413 

---

`DCO 1.1 Signed-off-by: Scott Nath<github@scottnath.com>`

### DIFF
--- a/lib/applications/routes.js
+++ b/lib/applications/routes.js
@@ -249,14 +249,12 @@ const save = (req, res) => {
       _.unset(req.session, 'form.applications.save');
 
       // remove this app from apps object
-      apps = apps.map(app => {
+      apps = apps.filter(app => {
         if (app.id === id) {
-          return null;
+          return false;
         }
 
-        return app;
-      }).filter(app => {
-        return app !== null;
+        return true;
       });
 
       // replace apps in the application's settings
--- a/lib/applications/routes.js
+++ b/lib/applications/routes.js
@@ -239,8 +239,8 @@ const secret = (req, res, next) => {
 const save = (req, res) => {
   const id = _.get(req.session, 'form.applications.edit.id', null);
   const referrer = (_.get(req.session, 'referrer') || req.get('Referrer')) || `/${config.applications.base}`;
-  const apps = req.app.get('applications-apps') || [];
   const merged = req.app.get('applications-merged') || {};
+  let apps = req.app.get('applications-apps') || [];
 
   // user hit delete button
   if (req.body.submit === config.applications.actions.delete && id) {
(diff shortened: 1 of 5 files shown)


lib/applications/routes.js
```diff
@@ -239,22 +239,28 @@ const secret = (req, res, next) => {
...
-      const index = apps.findIndex(app => {
-        return app.id === id;
+      // remove this app from apps object
+      apps = apps.map(app => {
+        if (app.id === id) {
+          return null;
+        }
+
+        return app;
+      }).filter(app => {
```
Snugug: Why not just filter and if  `app.id === id` return `false`?

lib/applications/send.js
```diff
@@ -0,0 +1,189 @@
...
+const save = options => {
+  const ids = options.endpoints.map(end => {
+    return end.id;
+  });
+
+  return database.transaction(trx => {
+    return trx('applications')
+      .select('id', 'responses')
+      .whereIn('id', ids).then(aps => {
+        return Promise.map(aps, app => {
```
Snugug: I'd like to see us start using the native Promise functionality instead of Bluebird as there are some oddities with Bluebird's implementations and we don't need the overhead to actually do this. For this, we do

```javascript
const apps = aps.map(THE PROMISE);
return Promise.all(apps);
```

lib/applications/send.js
```diff
@@ -0,0 +1,189 @@
...
+ */
+const save = options => {
+  const ids = options.endpoints.map(end => {
+    return end.id;
+  });
+
+  return database.transaction(trx => {
+    return trx('applications')
+      .select('id', 'responses')
+      .whereIn('id', ids).then(aps => {
```
Snugug: Why is this `then` not on the next line?

lib/applications/send.js
```diff
@@ -0,0 +1,189 @@
...
+            json: {
+              id: app['client-id'],
+              secret: app['client-secret'],
+            },
+          },
+        };
+      }
+
+      return null;
+    }).filter(Boolean);
```
Snugug: This is now the second way I've seen you filter out things that are `null` in this PR. Can we stick with a single method?

lib/applications/send.js
```diff
@@ -0,0 +1,189 @@
...
+
+/*
+ * Promise wrapper for Request
+ */
+const req = options => {
+  return new Promise((res, rej) => {
+    request(options, (err, response) => {
+      let status;
+      if (err) {
+        rej(err);
```
Snugug: I thought we had discussed not rejecting errors

lib/applications/send.js
Snugug: You actually do this in the previous function

lib/applications/send.js
```diff
@@ -0,0 +1,189 @@
...
+ * @param  {options} options - see options typedef
+ *
+ * @returns {promise}  - knex transactions
+ */
+const save = options => {
+  const ids = options.endpoints.map(end => {
+    return end.id;
+  });
+
+  return database.transaction(trx => {
```
Snugug: Why are we using `transaction` here? This is the first I've seen of it in our codebase

Snugug: http://knexjs.org/#Transactions

scottnath: Switching to using `transaction` here because of the possibly large number of database calls that could be required by this function. Transaction opens the connection to the db and holds it open until all transactions are complete or it hits an error.
I'm hoping this curbs some of the `.trx` errors we've been getting during travis builds.

lib/applications/send.js
scottnath: needs to map first -  different than the one in routes.js (which only filters now)

lib/applications/send.js
scottnath: we had. I am ashamed (and it is now fixed)

scottnath:
@Snugug all changes added to PR

lib/applications/send.js
Snugug: 👍

PR #499, opened 2016-10-16
Hotfix/linting

Snugug:
_something_ broke in Runner where it's not bringing in the linting correctly anymore. So, adding linting files in here.

---
Resolves https://travis-ci.org/punchcard-cms/punchcard/builds/167761996

### DIFF
--- a/tests/applications.js
+++ b/tests/applications.js
@@ -71,7 +71,7 @@ const appset = (find, changed) => {
 const reqObj = {
   method: 'GET',
   url: '/application',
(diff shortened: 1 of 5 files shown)


tests/applications.js
```diff
@@ -72,8 +72,8 @@ const reqObj = {
   method: 'GET',
   url: '/application',
   app: { // eslint-disable-line quote-props
```
scottnath: can't this comment go away now that they are all quoted?

Issue #408, opened 2016-09-14
references in apis


scottnath:
Get references in APIs

``` gherkin
Feature: API References
  As an application developer
  I want referenced content to be available through a single API call
  So that I can retrieve all associated content at once

  Scenario: Basic
    Given that a piece of content has a reference to another piece of content
     When I retrieve that piece of content from the API by its id (/api/types/:type/:id)
     Then the referenced content value will be replaced with information about that piece of content if the referenced content is live
      And it will include its ID
      And it will include its type's name, slug, and API Endpoint URL /api/types/:type
      And it will include its key
      And it will include its key_slug
      And it will include the API Endpoint URL to the full piece of content /api/types/:type/:id

  Scenario: Depth
    Given that a piece of content has a reference to another piece of content
     When I retrieve that piece of content from the API by its id (/api/types/:type/:id)
      And I want to retrieve the attributes of referenced content
      And I want to retrieve the attributes of any further referenced content
     Then I will be able to add a depth parameter to my request
      And depth can be a number starting at 0 of the number of nested references to retrieve
      And the referenced content value will be replaced with information about that piece of content if the referenced content is live
      And it will include its ID
      And it will include its type's name, slug, and API Endpoint URL /api/types/:type
      And it will include its key
      And it will include its key_slug
      And it will include the attributes of the referenced content
      And if there are references in the attributes and the depth hasn't been reached, that reference will be replaced in this same way
      And if there are references in the attributes and the depth has been reached, that reference will be replaced as a basic reference
```

---

**API References**
- [ ] Basic
- [ ] Depth

scottnath:
@Snugug 

`And it will include its type's name, slug, and URL`
what is URL? The api url for this type? `/api/types/:type`?
maybe a `definition` for URL would help?

`And it will include its key`
Is `key` the identifier? If so, can we use `identifier` instead of `key`?

Snugug:
* I'll update the URL
* The API was built prior to us settling on that terminology. If you'd like to change it, separate issue as that's a breaking change. Going to keep this as `key` as that's what's currently there

scottnath:
ok. #keynamingfail #whatever

![friday-rebecca-black](https://cloud.githubusercontent.com/assets/216931/19191673/6da5eb5c-8c71-11e6-8a5d-5da4e2421534.gif)

scottnath:
@Snugug 

questions:

1. Is there a default `depth`? Is that default `depth: 1`?
[...]

PR #500, opened 2016-10-17
Hotfix/api testing

scottnath:
Turns on and repairs testing for APIs so that I can begin #408

---
Related #408 


tests/apis.js
```diff
@@ -1,88 +1,54 @@
...
-for (let i = 0; i < 50; i++) {
-  const name = ipsum({
-    count: 3,
-    units: 'words',
-    format: 'plain',
+test.cb.before(t => {
+  database.init().then(() => {
+    database('live').insert(content).then(() => {
+      t.end();
+    }).catch(e => { // because we can't return in a before, this catch doesn't bubble out
```
Snugug: Can you put this on the next line instead of inline with the function initilization

Snugug: We should also fail with `e.stack` instead of `e` or `e.message` (below)

tests/apis.js
```diff
@@ -1,88 +1,54 @@
...
-    },
-  };
-
-  content[i] = item;
-}
+//////////////////////////////
+// Utils - attributes
+//////////////////////////////
+test('Utils: attributes', t => {
+  const item = Math.round(Math.random() * content.length);
```
Snugug: This may be a source of errors. If `Math.random() === 1`, then `item` will equal `length`, and the next line will fail because `content[lenght]` doesn't exist.

tests/apis.js
```diff
@@ -1,88 +1,54 @@
...
-  }).then(() => {
-    t.end();
-  }).catch(e => {
-    console.error(e);
-    t.end();
+  t.is(typeof attributes, 'object', 'Should return an object.');
+  Object.keys(attributes).map(key => {
+    const attr = key.split('-');
+    if (attr[attr.length - 1] === 'referencer') {
+      t.true(isUUID(attributes[key]), 'includes a uuid');
```
Snugug: So this test will be changing for https://github.com/punchcard-cms/punchcard/issues/408 right?

tests/fixtures/live/objects/database-mocks.js
```diff
@@ -0,0 +1,79 @@
+/*
```
Snugug: Why does this exist instead of using `fixtures/_utils.js`

tests/apis.js
scottnath: changing it to `const expected = cloneDeep(content[item - 1]);`

tests/apis.js
scottnath: Yes. there will be a whole set of tests for `attributes` function. I have to take into account repeatables in these tests too.

tests/apis.js
Snugug: That won't work either, because if `Math.random() === 0` you'll get `-1`. Need to scope it to `0 <= number <= length`

PR #515, opened 2016-10-25
Feature/follow

### LINKED ISSUE #509: Add Follow parameter to APIs
Because of the use of fuzzy search for things like `slug` and `key_slug`, it'd be useful to be able to pass a `follow` parameter to get the results back for a given item as if it was a call directly to that piece of content's ID endpoint

**Definitions**
* individual piece of content - Individual Piece of Content scenario as originally defined in #114, plus whatever current additions to that have been made
* fall through - Be applied as if they had been called directly

```gherkin
Feature: API Follow
  As an application developer
  I want to be able to get full content attributes back for queried content
  So that I gain access to their attributes without needing to make multiple calls

  Scenario: Content Attributes
    Given I have retrieved a listing of content
      And I want the full content for each of the items in the listing
     When I include the `follow` query parameter in my API call
      And that I set it to `true`
     Then each item in the listing of content should be returned as if it was a an individual piece of content
      And all query parameters passed in should fall through to the individual piece of content
```

---

**API Follow**
- [ ] Content Attributes

Snugug:
Adds Follow functionality to APIs that list content

---
Resolves #509

### DIFF
--- a/lib/routes/api.js
+++ b/lib/routes/api.js
@@ -50,11 +50,7 @@ const routes = application => {
   // TODO: Return error if no items are found
   app.get('/api/types/:type', (req, res, next) => {
     const type = app.get('content-types').find(item => {
-      if (item.id === req.params.type.toLowerCase()) {
-        return true;
-      }
-
-      return false;
+      return item.id === req.params.type.toLowerCase();
     });
 
     api.ofType(req.query, type).then(formatted => {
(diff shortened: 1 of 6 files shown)


lib/routes/api.js
```diff
@@ -49,7 +49,11 @@ const routes = application => {
   // TODO: Tests
   // TODO: Return error if no items are found
   app.get('/api/types/:type', (req, res, next) => {
-    api.ofType(req.query, req.params.type).then(formatted => {
+    const type = app.get('content-types').find(item => {
```
scottnath: use `req.content.types` to be consistent with [other routes](https://github.com/punchcard-cms/punchcard/blob/master/lib/routes/content.js#L42) and to make this easier to break out for testing later.

PR #516, opened 2016-10-25
Hotfix/follow

Snugug:
Quick Hotfix for https://github.com/punchcard-cms/punchcard/issues/509

> Program testing can be used to show the presence of bugs, but never to show their absence!
>
> ~ Edsger W. Dijkstra

So, the reason this wasn't caught in our tests is because of the lack of route testing (where this actually wound up failing).

---

### DIFF
--- a/tests/applications.js
+++ b/tests/applications.js
@@ -388,7 +388,7 @@ test('Send - bad urls', t => {
     t.true(Array.isArray(app.responses.sunset), 'includes live responses, which is an array');
 
     const sorted = _.sortBy(app.responses.sunset, 'timestamp');
-    t.is(sorted[sorted.length - 1].response, 400, 'includes endpoint response');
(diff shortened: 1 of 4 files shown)


tests/applications.js
```diff
@@ -380,7 +388,7 @@ test('Send - bad urls', t => {
     t.true(Array.isArray(app.responses.sunset), 'includes live responses, which is an array');
 
     const sorted = _.sortBy(app.responses.sunset, 'timestamp');
-    t.is(sorted[sorted.length - 1].response, 500, 'includes endpoint response');
+    t.is(sorted[sorted.length - 1].response, 400, 'includes endpoint response');
```
Snugug: This is consistently failing for me with a `400` on local, but `500` on Travis. I don't understand

PR #517, opened 2016-10-27
Feature/references api

### LINKED ISSUE #408: references in apis

scottnath:
PR for getting references in api.

---

Resolves #408


Snugug:
Running tests on actual endpoints locally, see the following problems:

If getting `/api`, `/api/types/:type`, or `/api/types/:type/:id`, when attaching `follow=true` as a query, I do not get `attributes` as expected, but rather get `meta.url`. Depth should _only_ be used for depth of references, not for follow. When adding `depth=1` I get `attributes` as expected. Depth is _only_ for references. See this `AND` from [the scenario](https://github.com/punchcard-cms/punchcard/issues/408#issue-176940688):

```gherkin
And depth can be a number starting at 0 of the number of nested references to retrieve
```

Snugug:
- Referenced content is not expanded; it's in the API as the UUID
- When trying to increase depth, it doesn't. `depth=1` results in same output as `depth=2` and `depth=3`, etc…

There are quire a few changes here, and presumably this works in the tests (haven't reviewed the tests yet) but it's not working in practice. Was something not connected up right?

tests/apis.js
```diff
@@ -66,16 +142,49 @@ test('Utils: attributes', t => {
     return typ.id === expected['type-slug'];
   });
 
-  const attributes = apiUtils.attributes(expected.value, model.attributes);
+  // because depth is checked during attributes, depth 1 and 0 are the same for the attributes function
+  const query = {
+    depth: 1,
```
Snugug: We need tests with both `depth` included and when `depth` isn't inlcuded

Snugug:
* References in repeatable don't appear to work. All repeatable references return as `null`

Snugug:
A brief check of the tests running locally seems to confirm that the internals of the code that are unit test tested appear to be working as expected (save the `depth=0` part) so I'm led to believe this is a connection issue somewhere from the unit tests to the actual integration

scottnath:
So, I've hit a roadblock. I think the first issue we have is that `content-types` may be merging the types incorrectly when a repeatable is present. If you look [at this section in content-types](https://github.com/punchcard-cms/content-types/blob/e977434a74e75ecff311c4b10527ef6faef3729a/lib/content-types/merge.js#L189) you'll see that it creates an array of inputs, equal to the `min` number for repeatables, in the final merged objects of content-types. So, a config like this:
```yaml
  - type: reference
    id: bar-reference
    name: Bar reference
    repeatable:
      min: 5
      max: 10
    inputs:
      reference:
        settings:
          contentType: services
```

 results in something like this when I give a `min` of 5 for a repeatable input:

```
{
        "name": "Bar reference",
        "description": "An input plugin with reference",
        "validation": {},
        "inputs": [
          {
            "reference": {
              "validation": {
                "function": "referenceValidation",
                "on": "change"
              },
              "type": "select",
              "reference": true,
              "label": "Bar reference",
              "options": [],
              "settings": {
                "view": "select",
                "placeholder": "-- Choose One --",
                "contentType": "services"
              },
              "id": "9ba6bfd3-8a59-48fe-8bc7-e2f15e89d8cc--0",
              "name": "bar-reference--reference--0"
            }
          },
          {
            "reference": {
              "validation": {
                "function": "referenceValidation",
                "on": "change"
              },
              "type": "select",
              "reference": true,
              "label": "Bar reference",
              "options": [],
              "settings": {
                "view": "select",
                "placeholder": "-- Choose One --",
                "contentType": "services"
              },
              "id": "9ba6bfd3-8a59-48fe-8bc7-e2f15e89d8cc--1",
              "name": "bar-reference--reference--1"
            }
          },
          {
            "reference": {
              "validation": {
                "function": "referenceValidation",
                "on": "change"
              },
              "type": "select",
              "reference": true,
              "label": "Bar reference",
              "options": [],
              "settings": {
                "view": "select",
                "placeholder": "-- Choose One --",
                "contentType": "services"
              },
              "id": "9ba6bfd3-8a59-48fe-8bc7-e2f15e89d8cc--2",
              "name": "bar-reference--reference--2"
            }
          },
          {
            "reference": {
              "validation": {
                "function": "referenceValidation",
                "on": "change"
              },
              "type": "select",
              "reference": true,
              "label": "Bar reference",
              "options": [],
              "settings": {
                "view": "select",
                "placeholder": "-- Choose One --",
                "contentType": "services"
              },
              "id": "9ba6bfd3-8a59-48fe-8bc7-e2f15e89d8cc--3",
              "name": "bar-reference--reference--3"
            }
          },
          {
            "reference": {
              "validation": {
                "function": "referenceValidation",
                "on": "change"
              },
              "type": "select",
              "reference": true,
              "label": "Bar reference",
              "options": [],
              "settings": {
                "view": "select",
                "placeholder": "-- Choose One --",
                "contentType": "services"
              },
              "id": "9ba6bfd3-8a59-48fe-8bc7-e2f15e89d8cc--4",
              "name": "bar-reference--reference--4"
            }
          }
        ],
        "html": "{% if reference.type == \"radio\" %}<ul>{% for option in reference.options %}<li><input type=\"{{reference.type}}\" name=\"{{reference.name}}\" id=\"{{reference.id}}--{{loop.index}}\" value=\"{{option.value}}\" {% if option.value == reference.value %}checked{% endif %}><label for=\"{{reference.id}}--{{loop.index}}\">{{option.label}}</label></li>{% endfor %}</ul>{% else %}<label for=\"{{reference.id}}\">{{reference.label}}</label><select id=\"{{reference.id}}\" name=\"{{reference.name}}\">{% for option in reference.options %}<option value=\"{{option.value}}\"{% if option.value == reference.value %}selected{% endif %}>{{option.label}}</option>{% endfor %}</select>{% endif %}",
        "id": "bar-reference",
        "type": "reference",
        "repeatable": {
          "min": 5,
          "max": 10
        }
      },
```

Question: is the the correct behavior?
My answer: I think not, but we'll have to do a breaking change, probably on both ct/punchcard to fix it.

Because of this, I need to fix this...or account for it in my code in this PR. This does not fix _all_ the issues, but it certainly accounts for some of them

Snugug:
https://github.com/punchcard-cms/punchcard/pull/517#issuecomment-257016690
This conversation is better reserved for the [Content Types Issue Queue](https://github.com/punchcard-cms/content-types/issues)

Snugug:
This functionality is still not quite right. `depth=0` and `depth=1` do the same thing. If no depth is included, or `depth=0`, then we should get the listing-like view for referenced content (with `meta`). Depth counts the nested depth of those `meta` links we wind up following _for references_ we expand from a listing view in to a `one` view. This means `depth=0` shouldn't expand any references from their listing view in to their one view, `depth=1` should expand references in the the top-level content from their listing view to their one view, `depth=2` should expand references in the first set of retrieved references from their listing view to their one view _as well as_ references from top-level content, etc…

This is happening for both list endpoints (`/api`, `/api/types/:type`) and individual content endpoints (`/api/types/:type/:id`)

This basically feels like an "off by one" error.

Depth 0 vs Depth 1

Depth 1 vs Depth 2

Snugug:
Another UX thing that appears wrong: if an input plugin is a reference and only has one input, the content is wrapped in that input as opposed to just being available. This is true for both full and listing view. Like all of our other items in our APIs, if an input plugin only has one input, the content should be available directly instead of underneath the input's name. It should only be wrapped in the input name if there are multiple inputs in an input plugin.


screen shot 2016-11-10 at 11 01 05 am
screen shot 2016-11-10 at 11 00 52 am

scottnath:
@snugug....ok, now we're into a repeatable ... I think what we want in the repeatable is to go from this:

```
"bar-reference": [
  {
    "reference": {
      "id": "cc6dda1b-12ac-4422-884d-8981276dbcf9",
      "type": {
        "name": "Services",
        "slug": "services",
        "url": "/api/types/services"
      },
      "key": "referrer 2",
      "key_slug": "referrer-2",
      "meta": {
        "url": "/api/types/services/cc6dda1b-12ac-4422-884d-8981276dbcf9"
      }
    }
  },
  {
    "reference": {
      "id": "cc6dda1b-12ac-4422-884d-8981276dbcf9",
      ...
    },
  },
],
```

to this:

```
"bar-reference": [
  {
    "id": "cc6dda1b-12ac-4422-884d-8981276dbcf9",
    "type": {
      "name": "Services",
      "slug": "services",
      "url": "/api/types/services"
    },
    "key": "referrer 2",
    "key_slug": "referrer-2",
    "meta": {
      "url": "/api/types/services/cc6dda1b-12ac-4422-884d-8981276dbcf9"
    }
  },
  {
    "id": "cc6dda1b-12ac-4422-884d-8981276dbc123",
    ...
  },
],
```

Let me poke at this for a time...

scottnath:
@Snugug adjusted api output and corresponding test. Please re-review.

PR #524, opened 2016-11-03
adding transaction to follow

### LINKED ISSUE #520: core dev: follow knex multi-call
The [api's follow function](https://github.com/punchcard-cms/punchcard/blob/master/lib/api/follow.js) is used by either [all](https://github.com/punchcard-cms/punchcard/blob/master/lib/api/follow.js) or [oftype](https://github.com/punchcard-cms/punchcard/blob/c2f4e85981c056dff25454272aec719077a5ebe6/lib/api/apis.js#L146) and then subsequently calls the [one function](https://github.com/punchcard-cms/punchcard/blob/c2f4e85981c056dff25454272aec719077a5ebe6/lib/api/utils.js#L249). These all use knex database calls. Best practice for a combined stack of database calls with knex is to use knex's [transaction feature](http://knexjs.org/#Transactions) which will run all the db calls as a single unit. This is already happening in `one`, but we should add this functionality to `follow`.

```gherkin
Feature: multiple database calls
  As a core developer
  I want to use core knex functions when calling the database
  So that I can make multiple database calls at once and still have a fast system

  Scenario: all types
    Given I am calling the Punchcard api to receive all types
      And that I have added `follow=true` as a parameter to my call
      And that Punchcard makes multiple database calls when getting all types
     When Punchcard makes a multi-request database call
     Then it should start those calls with a knex transaction query builder
      And it should pass the resulting `trx` through all subsequent functions
      And follow should be part of those functions
      And follow should accept a `trx` parameter
      And if the `trx` parameter is empty, it should use the database function instead
      And it should pass `trx` to the one function.

  Scenario: one type
    Given I am calling the Punchcard api to receive one content types
      And that I have added `follow=true` as a parameter to my call
      And that Punchcard makes multiple database calls when getting all content in a type
     When Punchcard makes a multi-request database call
     Then it should start those calls with a knex transaction query builder
      And it should pass the resulting `trx` through all subsequent functions
      And follow should be part of those functions
      And follow should accept a `trx` parameter
      And if the `trx` parameter is empty, it should use the database function instead
      And it should pass `trx` to the one function.
```

---

**multiple database calls**
- [ ] all types
- [ ] one type

scottnath:
Quick fix for adding transaction calls to `follow`, `all` and `ofType`

---
Resolves #520

### DIFF
--- a/lib/api/apis.js
+++ b/lib/api/apis.js
@@ -66,8 +66,7 @@ const all = (query, models) => {
   const organize = utils.organize(query);
 
   return database.transaction(trx => {
-
-    const results = database('live').select('*');
+    const results = trx('live').select('*');
 
     if (query.key) {
       results.where({
--- a/lib/api/apis.js
+++ b/lib/api/apis.js
@@ -65,31 +65,34 @@ const types = app => {
 const all = (query, models) => {
   const organize = utils.organize(query);
 
-  const results = database('live').select('*');
+  return database.transaction(trx => {
 
-  if (query.key) {
-    results.where({
-      key: query.key,
-    });
-  }
-  else if (query.key_slug) { // eslint-disable-line camelcase
-    results.where({
-      'key-slug': query.key_slug, // eslint-disable-line camelcase
-    });
-  }
+    const results = database('live').select('*');
 
-  return results.orderBy(organize.sort.by, organize.sort.dir).offset(organize.page.offset).limit(organize.page.limit).then(rows => {
-    return database('live').count('id').then(total => {
-      if (query.follow) {
-        return follow(query, rows, models, total, organize);
-      }
+    if (query.key) {
+      results.where({
+        key: query.key,
+      });
+    }
+    else if (query.key_slug) { // eslint-disable-line camelcase
+      results.where({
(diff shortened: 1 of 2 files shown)


lib/api/apis.js
```diff
@@ -65,31 +65,34 @@ const types = app => {
...
-    results.where({
-      key: query.key,
-    });
-  }
-  else if (query.key_slug) { // eslint-disable-line camelcase
-    results.where({
-      'key-slug': query.key_slug, // eslint-disable-line camelcase
-    });
-  }
+    const results = database('live').select('*');
```
Snugug: This should use `trx('live')`

lib/api/follow.js
```diff
@@ -14,30 +14,30 @@ const database = require('../database');
...
-        return model.id === row['type-slug'];
-      });
-
-      if (typeof type === 'undefined') {
-        return new Promise(res => {
-          res({});
-        });
-      }
+const follow = (query, rows, types, total, organize, trx) => {
+  const db = trx || database;
```
Snugug: This needs a little work. If `trx` isn't passed in, it needs to create a transaction.

PR #528, opened 2016-11-08
Feature/pluggable storage

### LINKED ISSUE #492: Code Dev: Plugable Storage
We want to be able to have different storage for different environments. Because of that, we need a common input/output with each storage solution so Punchcard won't need to know which storage solution is being used.

``` gherkin
Feature: Plugable file storage
  As a core developer
  I want file storage options with common, asynchronous inputs and outputs
  So that I can have multiple storage options with the same interface.

  Scenario: Input
    Given that I will have a file or files which need to be stored
     When Punchcard sends the file or files to be stored
     Then it should include details on the file(s)
      And the details should include each file's name
      And the details should include each file's type
      And the details should include each file's location

  Scenario: Output with error
    Given that I gave the storage plugin a file which will cause errors
      And that those errors will happen when the storage plugin attempts to receive the file
     When the storage plugin receives an error when trying to store a file
     Then it should return an error object
      And the object should include a human-readable error string

  Scenario: Output with success
    Given that I gave the storage plugin a file which causes no errors
      And because it causes no errors, it is stored by the storage option that storage plugin called
     When storage plugin responds on successful storage of a file
     Then it should return a success object
      And that object should include a file path
      And that file path should be relative to the storage plugin's route.

  Scenario: Punchcard use when success
    Given that a user has submited a piece of content
      And that the content has no validation errors for any attributes
      And that said content included a file to be saved
      And that Punchcard has sent the file to the storage plugin
     When the storage plugin is successful in storing the file
     Then Punchcard receives a success response from the storage plugin
      And said response should include the relative file path for the file
      And Punchcard should save the piece of content to the database
      And said content should include the relative path to this file.

  Scenario: Punchcard use when failure
    Given that a user has submited a piece of content
      And that the content has no validation errors for any attributes
      And that said content included a file to be saved
      And that Punchcard has sent the file to the storage plugin
     When the storage plugin encounters an error when attempting to store the file
     Then Punchcard receives an error from the storage plugin
      And it should treat this error as a validation error
      And it should not save the data from the form
      And it should return the user to the data entry form
      And the data entry form should display the error from the plugin
      And it should display the error next to the file upload input
      And the error should look the same as other validation errors.
```

**storage plugin**
- [x] Input
- [ ] Error
- [x] Success
- [x] Punchcard good
- [ ] Punchcard bad

Snugug:
Add Pluggable Storage

This is a WIP PR.

Currently complete:

* Saving a file
* File storage being pluggable

In Progress

* Saving to Database
* Retrieving from Database
* Errors

Not Yet Started

* Retrieving from Object Store

---
Resolves #492

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -140,5 +140,6 @@ module.exports = {
     type: 'fs',
     dest: 'public/files',
     settings: {},
+    public: '/files', // Can include {{dest}} for the dest path
   },
 };
--- a/config/default.js
+++ b/config/default.js
@@ -136,4 +136,9 @@ module.exports = {
       missing: 'Workflow \'%workflow\' for Content Type \'%type\' not found',
     },
   },
+  storage: {
+    type: 'fs',
+    dest: 'public/files',
+    settings: {},
+  },
 };
--- a/content-types/bar.yml
+++ b/content-types/bar.yml
@@ -31,3 +31,4 @@ attributes:
   - type: file
     id: file-upload
     name: Awesome Files
+    repeatable: true
--- a/content-types/bar.yml
+++ b/content-types/bar.yml
@@ -28,7 +28,3 @@ attributes:
         settings:
           contentType: services
           view: radio
-  - type: file
-    id: file-upload
-    name: Awesome Files
-    repeatable: true
--- a/content-types/bar.yml
+++ b/content-types/bar.yml
@@ -17,14 +17,17 @@ attributes:
     name: Bar reference
     repeatable: true
     inputs:
-      reference: 
-        settings: 
+      reference:
+        settings:
           contentType: services
   - type: reference
     id: foo-reference
     name: Foo reference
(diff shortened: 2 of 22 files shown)


Snugug:
Question: Should I restrict this PR to only the pluggable storage, and have a 2nd story/PR to actually hook it up (as that's a large chunk of this work) or should we do it all at once. Looking at you @scottnath

Snugug:
Tests are likely to fail, haven't actually finished refining this yet

scottnath:
@Snugug yes, please restrict.

Snugug:
Okay @scottnath, should be GTG now

Snugug:
HAHA wow. Ok. Blowing up this branch and putting everything back on.

Snugug:
Closed in favor of https://github.com/punchcard-cms/punchcard/pull/529

PR #529, opened 2016-11-09
Pluggable Storage

### LINKED ISSUE #492: Code Dev: Plugable Storage

Snugug:
Add Pluggable Storage

---
Resolves #492
Replaces #528

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -140,5 +140,6 @@ module.exports = {
     type: 'fs',
     dest: 'public/files',
     settings: {},
+    public: '/files', // Can include {{dest}} for the dest path
   },
 };
--- a/config/default.js
+++ b/config/default.js
@@ -136,4 +136,9 @@ module.exports = {
       missing: 'Workflow \'%workflow\' for Content Type \'%type\' not found',
     },
   },
+  storage: {
+    type: 'fs',
+    dest: 'public/files',
+    settings: {},
+  },
 };
--- a/lib/storage/put.js
+++ b/lib/storage/put.js
@@ -17,9 +17,20 @@ try {
   adapter = require(`vinyl-${config.storage.type}`); // eslint-disable-line global-require
 }
 catch (e) {
+  console.error(`Could not find Vinyl Adapter 'vinyl-${config.storage.type}'. Falling back to 'vinyl-fs'.`); // eslint-disable-line no-console
   adapter = vfs;
 }
 
+/*
+ * Punchcard File
+ *
+ * Custom Vinyl Type
+ *
+ * Adds UUID (uuid), field the file came from (field), file MIME type (type), and relative path from pluggable storage root (rel) to standard Vinyl object.
+ *
+ * READ - uuid
+ * READ/WRITE - field, type, rel
+ */
 class PunchcardFile extends Vinyl {
   constructor(options) {
     super(options);
@@ -69,7 +80,16 @@ class PunchcardFile extends Vinyl {
   }
 }
 
-const mapFiles = file => {
+/*
+ * File DJ
+ *
+ * Turns a file in to a PunchcardFile object
+ *
+ * @param {object} file - A File object as received from a multi-part form
+ *
+ * @returns {object} - A Promise that resolves to a new PunchcardFile, with UUID, Type, Path, and Rel available
+ */
+const fileDJ = file => {
   return new Promise((res, rej) => {
     fs.readFile(file.path, (err, data) => {
       if (err) {
@@ -93,6 +113,15 @@ const mapFiles = file => {
   });
 };
 
+/*
+ * Punchcard Stream
+ *
+ * Takes an array of Vinyl-compatible PunchcardFile objects, streams them through the chosen adapter, collects the results, and wraps the whole thing in a promise.
+ *
+ * @param {array} items - Array of PunchcardFile objects
+ *
+ * @returns {object} - Keys are file fields, values are the result of the stream transformation
+ */
 const punchcardStream = items => {
   return new Promise((res, rej) => {
     const files = {};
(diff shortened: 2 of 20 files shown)


tests/fixtures/files/index.js
```diff
@@ -0,0 +1,72 @@
...
+const uuid = require('uuid');
+const _ = require('lodash');
+const path = require('path');
+const fs = require('fs');
+
+const all = [
+  path.join(process.cwd(), '', 'storage.js'),
+  path.join(process.cwd(), '..', 'index.js'),
+  path.join(process.cwd(), '..', 'Gulpfile.js'),
+  path.join(process.cwd(), '../src/images', 'bee.svg'),
```
scottnath: Maybe not the `bee` image as a test file. Could def see that going away...

tests/storage.js
```diff
@@ -0,0 +1,69 @@
...
+    if (result === false) {
+      t.fail('No result for input item');
+    }
+
+    t.true(result.hasOwnProperty('name'), 'Result has a name');
+    t.true(result.hasOwnProperty('path'), 'Result has a path');
+    t.true(result.hasOwnProperty('type'), 'Result has a type');
+
+    t.is(result.name, expected.name, 'Result name is same as expected name');
+    t.is(result.type, expected.type, 'Result name is same as expected name');
```
scottnath: `Result type is same...`

lib/storage/put.js
```diff
@@ -0,0 +1,123 @@
...
+      this._rel = pth;
+    }
+  }
+
+  static isCustomProp(name) {
+    return super.isCustomProp(name) && builtInProps.indexOf(name) === -1;
+  }
+}
+
+const mapFiles = file => {
```
scottnath: jsdoc

lib/storage/put.js
```diff
@@ -0,0 +1,123 @@
...
+      f.type = file.type;
+      f.path = `/${f.uuid}${path.extname(file.originalFilename)}`;
+      f.rel = `${f.uuid}${path.extname(file.originalFilename)}`;
+
+      res(f);
+    });
+  });
+};
+
+const punchcardStream = items => {
```
scottnath: jsdoc

lib/storage/put.js
```diff
@@ -0,0 +1,123 @@
...
+      .on('error', e => {
+        rej(e);
+      })
+      .on('end', () => {
+        res(files);
+      });
+  });
+};
+
+const put = files => {
```
scottnath: jsdoc

tests/fixtures/files/index.js
```diff
@@ -0,0 +1,72 @@
...
+  path.join(process.cwd(), '../src/images', 'punchcard-image.svg'),
+  path.join(process.cwd(), '../src/images', 'punchcard-404.svg'),
+];
+
+const raw = () => {
```
scottnath: jsdoc

tests/fixtures/files/index.js
```diff
@@ -0,0 +1,72 @@
...
+    });
+
+    files.splice(counter, 1);
+  } while (files.length > 0 && Math.random() > 0.5);
+
+
+  return items;
+};
+
+const saved = () => {
```
scottnath: jay es dock

config/default.js
```diff
@@ -136,4 +136,10 @@ module.exports = {
       missing: 'Workflow \'%workflow\' for Content Type \'%type\' not found',
     },
   },
+  storage: {
+    type: 'fs',
```
Snugug: Easier said than done…
https://www.npmjs.com/search?q=vinyl

Snugug: This uses any [Vinyl Adapter](https://www.npmjs.com/package/vinyl#what-is-a-vinyl-adapter) that your heart can desire. There's no possible way we can document all the possibilities

scottnath: actually, you've started it just by having that link! Can link to that list, then explain a bit more about how to drop in one of them?

scottnath: Just promoting the idea of documenting this a little bit within this PR.

Snugug:
@scottnath your docs, good sir

Harley cares greatly about good documentation

PR #544, opened 2016-11-11
Feature/users middleware

### LINKED ISSUE #490: user management route middleware
Our user management routes all have common checks that need to be done on the various parameters that make up a user management url. These checks should be done within an express middleware.

``` gherkin
Feature: URL checks before route creation
  As a core developer
  I want user management urls to be confirmed for correctness before the route-specific code runs

  Scenario: Init
    Given that we want to have the user data model, merged with it's required input plugins, available in some routes
     When we initialize our application
     Then the application should determine the users data model
      And the model should be merged with input plugins via the content types module
      And it should add the merged model to the app settings
      And it should add the merged model to the request object.

  Scenario: URL-check Middleware
    Given that our system can perform url-parts checks before it triggers a route's code
    When a route path is checked for correctness
    Then that test should happen as middleware
      And that middleware should be triggered for every user management route
      And it should trigger an error route on url errors
      And it should continue to create the route when zero url errors are present.

  Scenario: Redirect to error page
    Given that a user has attempted to visit a url which is incorrectly formed
      And that the user needs to know what mistake has been made
    When a user management route's url creates an error
    Then the user should receive an error message
      And the message should be a path-part specific error message for the problem
      And it should show the user an error page
      And that error page should contain the path-part specific error message
      And that error page should contain a safe url for the user to visit to try again.
```

---

**URL checks**
- [x] Init
- [x] Middleware
- [x] Redirect to error page

scottnath:
Setting up middleware and init in preparation of upgrading the user routes.

---
Resolves #490


lib/users/middleware/utils.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
```
Snugug: Why have specifically middleware utils (and create a sub middleware folder while you're at it) instead of just having user utils?

lib/users/middleware/utils.js
```diff
@@ -0,0 +1,35 @@
...
+ */
+const id = (req) => {
+  return new Promise((res, rej) => {
+    // no id in the request
+    if (!req.params.hasOwnProperty('id')) {
+      res(true);
+    }
+
+    // this can be the add or save action instead of a uuid
+    if (req.params.id === config.users.actions.add || req.params.id === config.users.actions.save) {
```
Snugug: Didn't you write code like this for content routes? Is it possible to reuse that instead?

tests/users.js
```diff
@@ -1,5 +1,5 @@
 import test from 'ava';
-import users from '../lib/users';
+import users from '../lib/users/index';
```
Snugug: This shouldn't have changed. That's the point of calling it `index.js`

scottnath:
@Snugug all changes in. Please re-review.

PR #545, opened 2016-11-11
Required text label bug fix - CSS

### LINKED ISSUE #530: The required to save/publish text display issues
The required to save/publish flags have some display issues.

Issues:

* Label still has the class `.required--save` or `.required--publish` which extends the color bg to the label, which we don't want.
* Need to add some spacing between the label and the required flag.
* When displaying input groups that are required, the Required to Save/Publish text is doubling and displaying twice.

screen shot 2016-11-09 at 1 24 57 pm

screen shot 2016-11-09 at 1 25 03 pm

Example:
https://moriarty-blue.herokuapp.com/content/api-method/add

ineedsubstance:
CSS change for issue #530 to fix the spacing issue between the label and the required text mark.

---
Resolves #530

### DIFF
--- a/src/sass/partials/components/_forms.scss
+++ b/src/sass/partials/components/_forms.scss
@@ -183,10 +183,12 @@
   &--save {
     background-color: setting-get('primary support color');
     font-weight: setting-get('normal font weight');
+    margin-left: setting-get('rhythm') / 2;
   }
 
   &--publish {
     background-color: setting-get('secondary support color');
     font-weight: setting-get('normal font weight');
+    margin-left: setting-get('rhythm') / 2;
   }
 }


ineedsubstance:
This is only the CSS fix. To complete the bug issue I will also be submitting a PR in the content-types repo.

ineedsubstance:
Related PR - https://github.com/punchcard-cms/content-types/pull/137

Snugug:
The only fi for this should be in `content_types` that includes a physical space between the two. Why the margin change?

ineedsubstance:
@Snugug - if you look at the original bug, there is a spacing issue between the label and the required text. This didn't show up for me on the style guide because an extra space was added to the label text. I thought it made sense fix these all together, but if you want me to create a separate "bug" for this CSS change, I can.

screen shot 2016-11-09 at 1 24 57 pm

Snugug:
@ineedsubstance The bug isn't _just_ that there isn't a visual space, it's that there isn't a physical one (the space character). Adding a physical space (like there is  in the Style Guide) is the fix for both the visual indication of the bug (no visual space) and underlying issue (no actual space character)

ineedsubstance:
@Snugug - ok, I see. Just out of curiosity, why is it better to add a character space instead of a visual one?

Snugug:
Accessibility.

ineedsubstance:
@Snugug - Got it. Thanks. Fixing this in `content-types` instead and closing this PR.

Issue #168, opened 2016-07-28
Administrator: User Routes


scottnath:
Punchcard has a new cleaner, unit-tested way to do routing. The user-management section needs to be updated to conform to this improved routing solution.

``` gherkin
Feature: User routes to the future
  As an administrator
  I want administrate user's and their permissions
  So that I can control who may or may not be able to use my CMS and at what levels.

  Scenario: Match all possible routes
    Given the possible routes for user management are finite
     When updating routes
     Then then we should have a route for each possible action that needs a route
      And the routes should include viewing all users
      And the routes should include adding a user
      And the routes should include editing or deleting a user

  Scenario: View All
    Given that I have one or more users registered
     When I navigate to /users
     Then I can see a list of all of the users registered
      And I can see their emails
      And I can see their roles
      And I can see the date they were added
      And I will have a link to edit the user

  Scenario: Add New
    Given that I wish to register a user
     When I navigate to /users
     Then I click on a button that will allow me to add a new user
      And if there is a list of registered users, the button will be above them

  Scenario: Add
    Given that I know the URLs of endpoints to call on user actions
     When I navigate to /users/add
     Then I can add a new user
      And I can add a user's email
      And the user's email will be required and must be an email address
      And I can add a user's password
      And the user's password is required
      And I can select a role for the user
      And there should be a default role selected
      And there should be no way to leave role unselected.

  Scenario: View/Edit One
    Given that I have a user registered
     When I navigate to /users/:id
     Then I can see and edit the user's email
      And I can reset the password
      And I can select a different role for the user
      And I can delete the user.
```

---

**User routes**
- [ ] match all routes
- [x] view all
- [ ] add new
- [ ] add
- [ ] view/edit

PR #552, opened 2016-11-15
express/node application fixtures

scottnath:
Setting up test files and fixtures.

---
Relates to #168 


lib/routes/users.js
```diff
@@ -22,29 +22,13 @@ const database = require('../database');
  * @returns {object} - Configured Express Application
  */
 const routes = application => {
-  return new Promise(resolve => {
-    const app = application;
+  const app = application;
 
+  return new Promise(resolve => {
```
Snugug: Saw this all over the place at Chrome Dev Summit. Whenever you just need to resolve as a promise with a single value, we can do [`Promise.resolve()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve). This might seriously simplify a bunch of our code. Not something we _need_ to do, but a thought.

tests/users.js
```diff
@@ -1,23 +1,63 @@
...
+//////////////////////////////
+test.cb.before(t => {
+  database.init().then(() => {
+    database('users').del().then(() => {
+      database('users').insert(dbmocks.rows).then(() => {
+        t.end();
+      });
+    });
+  }).catch(e => {
+    t.fail(e.message);
```
Snugug: `e.message` instead of `e.stack`?

scottnath:
@Snugug replaced with `e.stack`

PR #555, opened 2016-11-15
New component for My Latest content

### LINKED ISSUE #538: Core Developer: My current work component
Design and build a component to show all current work for that user.

``` gherkin
Feature: Access to content
  As a Core Developer,
  I want to provide the user with a list of the current content they are working on
  So that they will be able to find and access this content with ease.

  Scenario: My current work
    Given that, a user is looking for content that they are currently working on
     When they visit the CMS home page
     Then they will see a list of all the content they are the author of listed in reverse chronological order by last edited date
       And that list will display a maximum of 5 items
       And that list will include a link to view all items if there are more than 5
       And it will display the identifier
       And it will display the date it was last edited
       And it will display a link that will allow users to edit the latest revision
       And it will display the status of that content
```
---

**Access to content:**
- [x] My current work

ineedsubstance:
New component for My Latest content that is a table and will live on the dashboard. This component can really be used for all tabular data moving forward.

---
Resolves #538 


ineedsubstance:
Had some challenges getting the [data](https://github.com/punchcard-cms/punchcard/pull/555/files#diff-d28313b4f2a6fbcc0515e9782942b4f4) setup correctly for the time being. Maybe this is a moot point since we will be importing real data eventually.

views/components/tabular-data.html
```diff
@@ -0,0 +1,29 @@
+{% macro tabularData(myLatest) %}
```
Snugug: `myLatest` is not a great name for the content being used. We want to use as few words as possible, as clear as possible, with as little camelCase as possible. As the macro is going to be called `table`, its argument should be something like `data`

views/components/tabular-data.html
```diff
@@ -0,0 +1,29 @@
...
+<article class="base--article" itemscope itemtype="http://schema.org/Table">
+  <header class="base--header">
+    <h1 class="base--h1" itemprop="name">{{myLatest.title}}</h1>
```
Snugug: Why does the table have both a header with a `h1` _and_ a `caption`? A `caption` is the title of a table.

views/components/tabular-data.html
```diff
@@ -0,0 +1,29 @@
...
+  </header>
+  <table class="base--table">
+    {% if myLatest.caption %}
+    <caption class="base--caption" itemprop="caption" itemscope itemtype="https://schema.org/Property">{{myLatest.caption}}</caption>
+    {% endif %}
+    <thead class="base--thead">
+      <tr class="base--tr">
+        {% for tableHeader in myLatest.tableHeaders %}
```
Snugug: `for header in data.headers`

views/components/tabular-data.html
Snugug: Why does the table have both a header with a `h1` _and_ a `caption`? A `caption` is the title of a table.

views/components/tabular-data.html
```diff
@@ -0,0 +1,29 @@
...
+        <th class="base--th">{{tableHeader}}</th>
+        {% endfor %}
+      </tr>
+    </thead>
+    <tbody class="base--tbody">
+        {% for table in myLatest.table %}
```
Snugug: `for row in data.body`

views/components/tabular-data.html
```diff
@@ -0,0 +1,29 @@
...
+        <tr class="base--tr">
+          <td data-th="{{table.headers[0]}}" class="base--td" itemprop="name" itemscope itemtype="https://schema.org/Table">{{table.identifier}}</td>
```
Snugug: All of these `td`s shouldn't have an `itemtype` or `itemscope`. The `table` itself should have them, but nothing underneath. If the whole row is going to have a different `itemscope`, with `itemprop`s different than `Table`, that should be put on the `tr`

Snugug: You seem to be duplicating header information here. The `data-th` should be drawing from the same item generating the `th`s (`data.headers`, as I'm recommending). This also hardcodes the number table data cells in a row, regardless of number of headers. The number of cells should be either dynamically generated based on the number of headers there are, or it should be dynamically generated based on an array of values.

I'd recommend something like the following:

```html
{% for row in data.body %}
<tr class="base--tr">
  {% for header in data.headers %}
  <td data-th="{{header}}" class="base--td">{{row[loop.index]}}</td>
  {% endfor %}
</tr>
{% endfor %}
```

`data.body` should be an array of values

views/styles.html
```diff
@@ -1,12 +1,16 @@
 {% extends "_donut.html" %}
 
-{% block pageTitle %}{{pageTitle}}{% endblock %}
-
 {% block pageHead %}
   {{ super() }}
 {% endblock %}
 
 {% block main %}
+  <section class="_style-guide--group">
```
Snugug: While I see what you're looking to do here, the Punchcard style guide isn't set up in the same way that our previous work was. Punchcard needs to be updated to do so in order to use this (please do so).

In addition, custom components should be below our base styling

lib/routes/data/styles.js
```diff
@@ -0,0 +1,86 @@
+'use strict';
+
+/*
+ * Sample Data for Styles Route
+ *
+ * @returns {object} - Object containing sample data to use when rendering samples in the Style Guide
+ */
+
+module.exports = {
+  myLatest: {
```
Snugug: As we're now just starting to build out a proper style guide, I would like our data to follow the following semantics:

```javascript
componentData: [ // name of the component, plus Data
  {
    name: 'Sample component instance name', // when we display multiple versions of a component, we need subtitles for the components
    data: {}, // The actual data for use in this component
  },
];
```

Then, in the Style Guide generation, we'll do something like the following:

```html
<h2 class="_style-guide--title">Component</h2>
  {% from "components/component.html" import component %}
  {% for example in componentData %}
  <h3 class="_style-guide--instance>{{example.name}}</h3>
  {{component(example.data)}}
</section>

Comment (ineedsubstance) [inline lib/routes/data/styles.js]: I tried to match your example, but had issues getting it to work.

File: views/components/tabular-data.html (hunk 5 above)
Comment (ineedsubstance) [inline views/components/tabular-data.html]: Tried this, but the `{{row[loop.index]}}` didn't work for me and I think how I have the `header` data setup that wouldn't' work either. Basically, I couldn't get the `headers` to show up here.

File: views/styles.html (hunk 6 above)
Comment (ineedsubstance) [inline views/styles.html]: Made some changes to try and match what we had on `wox-ui` but we will still need to create stories to take the existing typography stuff and move that into components like @poofichu already did on `wox-ui`.

File: views/components/table.html [hunk 8]
```diff
@@ -0,0 +1,25 @@
+{% macro table(tableData) %}
```
Snugug: Should be `data` not `tableData`. We're in the context of a table

views/components/table.html
```diff
@@ -0,0 +1,25 @@
...
+<article class="base--article" itemscope itemtype="http://schema.org/Table">
+  <header class="base--header">
+    <h1 class="base--h1" itemprop="name">{{tableData.name}}</h1>
```
Snugug: The h1 should have been dropped in favor of `caption`. A `caption` is the title of a table.
http://html5doctor.com/element-index/#caption

As this is a simple table component, it also shouldn't be wrapped in an `article`. `table` is a self-contained semantic object

views/components/table.html
```diff
@@ -0,0 +1,25 @@
...
+  </header>
+  <table class="base--table">
+    <thead class="base--thead">
+      <tr class="base--tr">
+        {% for header in tableData.data.headers %}
```
Snugug: The header information should live in `data.headers`, `data` coming from the passed in `data` in the macro. This should read `{% for header in data.headers %}`

views/components/table.html
```diff
@@ -0,0 +1,25 @@
...
+        <th class="base--th">{{header}}</th>
+        {% endfor %}
+      </tr>
+    </thead>
+    <tbody class="base--tbody">
+        {% for row in tableData.data.row %}
```
Snugug: Please try https://github.com/punchcard-cms/punchcard/pull/555#discussion_r88254977 with the correct data model as described above.

lib/routes/data/styles.js
```diff
@@ -0,0 +1,63 @@
...
+/*
+ * Sample Data for Styles Route
+ *
+ * @returns {object} - Object containing sample data to use when rendering samples in the Style Guide
+ */
+
+module.exports = {
+  tableData: {
+    name: 'My Latest',
+    data: {
```
Snugug: Should just be `headers`

lib/routes/data/styles.js
```diff
@@ -0,0 +1,63 @@
...
+      headers: [
+        'content',
+        'last edited',
+        'actions',
+        ' ',
+      ],
+      row: [
```
Snugug: Should be `rows` (or in my previous comments, `body`) and it should be an array of arrays, not an array of objects that each contain one key that's an array.

```javascript
rows: [
  [
    'Conversation API',
    'Tuesday, November 15',
    'Edit Latest Revision',
    'Draft',
  ],
  [
    'Conversation Overview',
    'Tuesday, November 16',
    'Edit Latest Revision',
    'Draft',
  ],
]
```

views/styles.html
```diff
@@ -331,5 +327,13 @@ <h1 class="base--h2">Form elements</h1>
       </div>
     </form>
   </section>
-</main>
+
+  <h2 class="_style-guide--title">Tabular Data</h2>
```
Snugug: Headers for a style guide group belong inside that group's section.

views/components/table.html
```diff
@@ -0,0 +1,25 @@
...
+        <tr class="base--tr" itemprop="dataset" itemscope itemtype="https://schema.org/Dataset">
```
Snugug: This is not a [dataset](https://www.w3.org/wiki/WebSchemas/Datasets). A dataset [is differentiated because](https://en.wikipedia.org/wiki/Data_set) "where every column of the table represents a particular variable, and each row corresponds to a given member of the data set in question"

Not all tables are datasets (and the key thing here is that Dataset is for a whole table, not rows in a table). Datasets you can think of almost as spec sheets. This content, on the other hand, is rather a tabular view of content and actions that can be taken When thinking dataset, consider the [demos provided by WebSchemas](https://logd.tw.rpi.edu/schemaorg_dataset_extension).

Snugug: In addition, `dataset` [is not a valid Table `itemprop`](http://schema.org/Table)

Snugug:
As for the links problem, you can do something like the following in your template:

```handlebars
{% for row in data.rows %}
<tr class="base--tr">
  {% for header in data.headers %}
  {% set r = row[loop.index] %}
  <td data-th="{{header}}" class="base--td">{% if r.url and r.title %}<a href="{{row.url}}">{{r.title}}</a>{% else %}{{r}}{% endif %}</td>
  {% endfor %}
</tr>
{% endfor %}
```

This will check to see if the row is an object that contains the `url` and `title` keys and, if it does, prints out a link instead of printing out the static content

Snugug:
Here is what the final data model and HTML should look like. Please look it over and ask any questions (doing this instead of committing directly so you can see the content)

**Data Model**
```javascript
'use strict';

/*
 * Sample Data for Styles Route
 *
 * @returns {object} - Object containing sample data to use when rendering samples in the Style Guide
 */

module.exports = {
  tableData: {
    name: 'My Latest',
    headers: [
      'content',
      'last edited',
      'actions',
      ' ',
    ],
    body: [
      [
        'Conversation API',
        'Tue Nov 15 2016 11:09:03 GMT+0000 (UTC)',
        'Edit Latest Revision',
        {
          url: '/drafts',
          title: 'Draft',
        },
      ],
      [
        'Conversation Overview',
        'Mon Nov 14 2016 03:22:12 GMT+0000 (UTC)',
        'Edit Latest Revision',
        'Draft',
      ],
      [
        'Python SDK',
        'Fri Nov 11 2016 12:12:01 GMT+0000 (UTC)',
        'Edit Latest Revision',
        'Draft',
      ],
      [
        'News Intelligence',
        'Wed Oct 26 2016 11:31:26 GMT+0000 (UTC)',
        'Edit Latest Revision',
        'Approved',
      ],
      [
        'Cognitive Head Hunter',
        'Wed Oct 26 2016 18:42:40 GMT+0000 (UTC)',
        'Edit Latest Revision',
        'Needs Approval',
      ],
    ],
  },
};
```

**Nunjucks**
```handlebars
{% macro table(data) %}
<table class="base--table">
  <caption class="base--h1">{{data.name}}</caption>
  <thead class="base--thead">
    <tr class="base--tr">
      {% for header in data.headers %}
      <th class="base--th">{{header}}</th>
      {% endfor %}
    </tr>
  </thead>
  <tbody class="base--tbody">
    {% for rows in data.body %}
    <tr class="base--tr" itemprop="dataset" itemscope itemtype="https://schema.org/Dataset">
      {% for header in data.headers %}
      {% set row = rows[loop.index0] %}
      <td data-th="{{header}}" class="base--td">{% if row.url and row.title %}<a href="{{row.url}}">{{row.title}}</a>{% else %}{{row}}{% endif %}</td>
      {% endfor %}
    </tr>
    {% endfor %}
  </tbody>
</table>
{% endmacro %}
```

The caption is going to need custom styling (for instance, by default it's center aligned, and there's probably some spacing we're going to want to add). We're also likely going to want to check to see if there's a name or not before adding a caption.

lib/routes/data/styles.js
ineedsubstance: ok

lib/routes/data/styles.js
ineedsubstance: ok, will try this again

views/components/table.html
ineedsubstance: updated

views/components/table.html
ineedsubstance: ok

views/components/table.html
ineedsubstance: ok, will take a look

ineedsubstance: I'll dig deeper on this

views/styles.html
ineedsubstance: updated

views/components/table.html
ineedsubstance: ok, I think I understand what this shouldn't be, but I can't figure out what it should be. I moved the `itemscope` and `itemtype` to the `table` level and removed `itemprop` for now since I'm not sure what that should be.

ineedsubstance:
@Snugug - got everything to show up thanks to your recommendations. Still had some [trouble/questions](https://github.com/punchcard-cms/punchcard/pull/555/files/af2252e1324d151539e8f89d0bf5efcd6eb34d07#diff-4692d3a94d9649e71080277c0364d577) about the schema stuff.

Also, my build is failing, but I'm not sure why. Tests run fine locally.

ineedsubstance:
@Snugug - I think we are good to go on this.

views/components/table.html
```diff
@@ -0,0 +1,22 @@
...
+  <thead class="base--thead">
+    <tr class="base--tr">
+      {% for header in data.headers %}
+      <th class="base--th">{{header}}</th>
+      {% endfor %}
+    </tr>
+  </thead>
+  <tbody class="base--tbody">
+    {% for rows in data.body %}
+    <tr class="base--tr" itemprop="dataset" itemscope itemtype="https://schema.org/Dataset">
```
Snugug: This was an error on my part. This should not have its own `itemscope` or an `itemprop` or `itemtype`

views/components/table.html
```diff
@@ -0,0 +1,22 @@
+{% macro table(data) %}
+<table class="base--table">
+  <caption class="base--caption">{{data.name}}</caption>
```
Snugug: As mentioned in my comments in https://github.com/punchcard-cms/punchcard/pull/555#issuecomment-261702309, this should be wrapped in an `if` to check to see if `data.name` exists, and only if it does should the caption be added

ineedsubstance: my bad, I missed that. will fix.

views/components/table.html
ineedsubstance: ok, will remove

Issue #317, opened 2016-08-26
First time setup, path to create an admin user


mbarlock:
After successfully setting up punchcard there is no default admin account. There should be a path or initial setup step that will create an admin user

``` gherkin
Feature: Prompting to create an admin
  As a developer standing up punchcard
  I want to be prompted to create the first admin user
  So that I can log in and edit content without modifying the database directly

  Scenario: Developer stands up punchcard and goes to the root
    Given a brand new instance of punchcard
     When the user hits any route
     Then the user is prompted to create a new admin account
```

---

**Prompting to create an admin**
- [ ] Developer stands up punchcard and goes to the root

scottnath:
Hi @mbarlock, this is a known issue, but we don't have an issue for it written yet. Can you please submit a bug report according to our [contributing guidelines](https://github.com/punchcard-cms/punchcard/blob/master/CONTRIBUTING.md#issue-guidelines)?

In the meantime, run the unit tests after `npm install` by running `npm test`. When you do that, it creates a test admin user with [these credentials](https://github.com/punchcard-cms/punchcard/blob/master/tests/server.js#L17) for test purposes...but you can use that user to log in and create more users, including admins.

Snugug:
As this is a new feature, not a bug, please follow our contributing guidelines for requesting new functionality

mbarlock:
Updated! Took my best stab, at writing out the gherkin, leme know if I can do better

Snugug:
My feeling from a technical implementation perspective is that when a user is redirected to login, we check to see if if there are any users in the database, and if there are not, we change that form to a create user form

scottnath:
Sounds good.

Working on route middleware upgrades now.

scottnath:
This relies on the upgrades to the user's routing and model

scottnath:
blocked by #560

scottnath:
@mbarlock @Snugug 
fix is in!
https://github.com/punchcard-cms/punchcard/pull/572

PR #560, opened 2016-11-16
Feature/user routes

### LINKED ISSUE #168: Administrator: User Routes

scottnath:
Modernizing the User routes.

---
Resolves #168 


lib/users/routes.js
```diff
@@ -30,18 +35,203 @@ const all = (req, res, next) => {
...
+ * @param {object} res - HTTP Response
+ * @param {object} next - Express callback
+ *
+ * @returns {promise} content-types promise
  */
-const add = () => {};
+const add = (req, res, next) => {
+  const errors = _.get(req.session, 'form.users.save.errors', {});
+  const values = _.get(req.session, 'form.users.save.content', {});
+  const merged = req.app.get('users-model') || {};
```
Snugug: On user add pages, password is required, but in the actual implementation it is not

lib/users/routes.js
```diff
@@ -30,18 +35,203 @@ const all = (req, res, next) => {
...
+
+  if (validated === true) {
+    const data = {
+      'email': req.body['email--email'],
+      'role': req.body['role--select'],
+      'password': bcrypt.hashSync(req.body['password--password']),
+    };
+
+    // `update` is for the Edit form
+    if (req.body.submit === config.users.actions.update && id) {
```
Snugug: On update of users, if `password` field is left blank, it shouldn't overwrite the password in the database. It should only do so if `password` is changed.

lib/users/routes.js
```diff
@@ -30,18 +35,203 @@ const all = (req, res, next) => {
...
+ * @param {object} next - Express callback
+ *
+ * @returns {promise} resolves/rejects promise
  */
-const one = () => {};
+const one = (req, res, next) => {
+  const id = _.get(req.session, 'form.users.edit.id', null);
+  const errors = _.get(req.session, 'form.users.save.errors', {});
+  const values = _.get(req.session, 'form.users.save.content', {});
+  const merged = req.app.get('users-model') || {};
```
Snugug: The description for password on a user edit form should explain how entering a new password will change the existing one

lib/users/routes.js
```diff
@@ -30,18 +35,203 @@ const all = (req, res, next) => {
...
+      _.unset(req.session, 'form.users.save');
+
+      res.redirect(`/${config.users.base}`);
+
+      return true;
+    });
+  }
+
+  // Validation
+  const validated = types.form.validate(req.body, merged, 'save');
```
Snugug: Validation, on a new user, needs to check to ensure the email address entered isn't already in use by another user

lib/users/routes.js
```diff
@@ -30,18 +35,203 @@ const all = (req, res, next) => {
...
+        res.redirect(`/${config.users.base}`);
+
+        return true;
+      });
+    }
+
+    return database('users').insert(data).returning('*').then(added => {
+      _.unset(req.session, 'form.users.save');
+
+      res.redirect(`/${config.users.base}/${added[0].id}`);
```
Snugug: Why does creating a new user redirect to the individual user's page instead of the user listing page? There is no further action to take on the individual user page.

views/users/all.html
```diff
@@ -30,8 +30,7 @@ <h1 class="introduction--h1">{{title|capitalize}}</h1>
         <td data-th="Email" class="base--td">{{user.email}}</td>
         <td data-th="Role" class="base--td">{{user.role}}</td>
         <td data-th="Date" class="base--td">{{user.created}}</td>
-        <td data-th="Actions" class="base--td"><a href="{{config.base}}/{{user.id}}/{{config.actions.edit}}" class="base--link">{{config.actions.edit|capitalize}}</a></td>
-        <td data-th="" class="base--td"><a href="{{config.base}}/{{user.id}}/{{config.actions.delete}}" class="base--link">{{config.actions.delete|capitalize}}</a></td>
+        <td data-th="Actions" class="base--td"><a href="{{config.base}}/{{user.id}}" class="base--link">{{config.actions.edit|capitalize}} / {{config.actions.delete|capitalize}}</a></td>
```
Snugug: `<a href="/{{config.base}}`

lib/users/routes.js
scottnath: This is actually a mistake I made when writing the story. In the current implementation, password is not required in either situation. Adding a required password _should_ be out of scope for this story. Reasoning:

* Passwords are not currently required
* User forms are created using `content-types` and we would have to do some serious manipulation of the `merged` content-type object to make this have a required password in one situation and not have passwords required in another

I think that in the [first time setup story](https://github.com/punchcard-cms/punchcard/issues/317) the solution will be to have a different form for first time users that requires a password. This form can then _also_ be used when adding a new user.

request: I can skip your `required` password comments in this PR and defer to the first-time user story to add that functionality as that is _new_ and should not be part of this refactor

lib/users/routes.js
scottnath: added

lib/users/routes.js
scottnath: agreed. fixed in latest

lib/users/model.js
```diff
@@ -36,13 +36,14 @@ const structure = {
...
       name: 'Email',
+      required: 'save',
       description: 'Email address for your account',
     },
     {
       type: 'password',
       id: 'password',
       name: 'Password',
-      description: 'Gotta get in',
+      description: 'CAUTION: Adding a password here will change the user\'s password.',
```
Snugug: This should only be true on edit forms, not on new user forms

Snugug: No, that's fine. Just trying to do code review based on the provided story. This code doesn't fully resolve the story it's for, but I'll merge it given that there is another story in progress that will encompass the required changes

Snugug:
@scottnath you need to update the `yarn.lock` file as well or this won't actually downgrade Yarn

package.json
```diff
@@ -103,7 +103,7 @@
     "mock-express-response": "^0.1.2",
     "nock": "^9.0.0",
     "node-mocks-http": "^1.5.3",
-    "nyc": "^9.0.1",
+    "nyc": "^6.6.0",
```
Snugug: Tou need to update the `yarn.lock` file as well or this won't actually downgrade Yarn

scottnath:
@Snugug all tests pass now :/

PR #571, opened 2016-11-17
Add button styles to add and delete

### LINKED ISSUE #556: Bug:  Add Button Needs Visual Styling
The add button needs some visual love attached to it.  Looks like it was given the class `form--add`, so that class needs to extend the button mixin.

image

poofichu:
Closes #556 

image

### DIFF
--- a/src/sass/partials/components/_forms.scss
+++ b/src/sass/partials/components/_forms.scss
@@ -24,6 +24,19 @@
     }
   }
 
+  &--add {
+    @include button(setting-get('primary accent color'), setting-get('primary accent color'), setting-get('primary color'), setting-get('primary accent color'), setting-get('primary color'), setting-get('primary color'), setting-get('primary color'));
+    margin-top: setting-get('rhythm') * 2;
+  }
+
+  &--delete {
+    @include button(color('neutral-white', 40), setting-get('primary color'), setting-get('primary color'), color('neutral-white', 40), setting-get('primary color'), setting-get('primary color'), setting-get('primary color'));
+  }
+
+  &--repeatable {
+    margin-top: setting-get('rhythm') * 2;
+  }
+
   &--description {
     color: color('grey', 60);
     font-size: ms(-1);


Snugug:
@poofichu Styling thoughts for you:

* Do we want `delete` to be that prominent, or do we want it to be similar to the `cancel` we have where it appears as a subdued link?
* Do we want to keep the final `delete` and `add` on separate lines, or line them up together?
* Is the placement for `delete` where we want it to be?

poofichu:
@ineedsubstance thoughts on the placement/button style of `delete`?

image

ineedsubstance:
@poofichu - I like this version where the `delete` action is secondary and the `add` is primary. Placement looks good. I do wish we could explore (maybe in the future), adding icons to the primary buttons. So in this case, adding a `+` icon to that button could be a nice addition. But overall, I like this approach.

poofichu:
@ineedsubstance Ask and you shall receive:

image

ineedsubstance:
@poofichu - Wow, nice! I like it. Makes it stand out a little more and puts more emphasis on the purpose of the button. As long as you are cool with it, let's roll with it.

src/images/add.svg
```diff
@@ -0,0 +1,4 @@
+<svg id="Layer_1" data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 9.5 9.5">
```
Snugug: Can we put this in `src/images/icons` please

Snugug: And add all of the icon classes to this, please

src/sass/partials/components/_forms.scss
```diff
@@ -24,6 +24,34 @@
     }
   }
 
+  &--add {
+    @include button(setting-get('primary accent color'), setting-get('primary accent color'), setting-get('primary color'), setting-get('primary accent color'), setting-get('primary color'), setting-get('primary color'), setting-get('primary color'));
+
+    background-image: url('/images/add.svg');
+    background-position: setting-get('rhythm') 49%;
+    background-repeat: no-repeat;
+    background-size: 9px;
```
Snugug: Background size in `em` instead of `px` please. Generally don't use `px`

src/sass/partials/components/_forms.scss
```diff
@@ -24,6 +24,34 @@
     }
   }
 
+  &--add {
+    @include button(setting-get('primary accent color'), setting-get('primary accent color'), setting-get('primary color'), setting-get('primary accent color'), setting-get('primary color'), setting-get('primary color'), setting-get('primary color'));
+
+    background-image: url('/images/add.svg');
+    background-position: setting-get('rhythm') 49%;
```
Snugug: Why is this `49%`? Why not `50%`? If it's because it doesn't vertically or horizontally center properly, use `calc` to nudge it in to the right position

src/sass/partials/components/_forms.scss
```diff
@@ -24,6 +24,34 @@
     }
   }
 
+  &--add {
+    @include button(setting-get('primary accent color'), setting-get('primary accent color'), setting-get('primary color'), setting-get('primary accent color'), setting-get('primary color'), setting-get('primary color'), setting-get('primary color'));
+
+    background-image: url('/images/add.svg');
```
Snugug: We aren't going to be able to guarantee that this background image is always going to be there. This should be an inline SVG in the HTML

Snugug: If you have problems with that, extend https://github.com/punchcard-cms/punchcard/blob/master/lib/init/views.js like we did before

scottnath:
@poofichu are changes forthcoming this day, the day of monday, the day after a holiday, changes shall come forth, so be it and so it is said?

poofichu:
@scottnath going to work on this today!  Problem we are having is how to inject the svg code for a `plus` icon for the add button instead of using it as a background image.

src/sass/partials/components/_forms.scss
poofichu: @Snugug So if we're going to add the `svg` to the `html` I need to do so in this file, yah?
https://github.com/punchcard-cms/content-types/blob/master/lib/form/html.js#L179

Can you do nunjucks partials in `js` files?

src/sass/partials/components/_forms.scss
```diff
@@ -24,6 +24,34 @@
...
+    margin-top: setting-get('rhythm') * 2;
+    padding-left: setting-get('rhythm') * 2.5;
+  }
+
+  &--delete {
+    @include link(setting-get('primary color'), setting-get('primary color'), setting-get('secondary link color'), setting-get('secondary link color'));
+
+    background-color: transparent;
+    border: {
+      left: 0 none;
```
Snugug: If the width of all of these is `0` then they don't need their style set to `none`

src/sass/partials/components/_forms.scss
```diff
@@ -24,6 +24,34 @@
     }
   }
 
+  &--add {
+    @include button(setting-get('primary accent color'), setting-get('primary accent color'), setting-get('primary color'), setting-get('primary accent color'), setting-get('primary color'), setting-get('primary color'), setting-get('primary color'));
+
+    background-image: url('/images/icons/add.svg');
```
Snugug: This may be easier for us to read if `image`, `position`, `repeat`, and `size` were all nested under `background` like you're doing with `border` below

Issue #163, opened 2016-07-28
Administrator: Access user admin


scottnath:
Users need to be limited in the ability access to administrate users

```gherkin
Feature: User admin limited access
  As an administrator
  I want a limited access to user administration
  So that I can control which users can access user administration

  Scenario: Admins can update users
    Given that administrators need to update users
      And that access to changing users should not be allowed for non-admins
     When a user tries to access the user-admin interface
     Then that user is allowed access if they are an admin
      And that user is not allowed access if they are not an admin

  Scenario: Other roles can access user admin
    Given that we need roles to vary in what they may be able to access
    When a user tries to access the user-admin interface
    Then the system should check that user's role's allowed access
      And it should grant access if their role allows it
      And it should deny access if their role does not allow it.
```

---

**User admin access**
- [ ] admins
- [ ] other roles

mkiang3051:
I have a couple of questions that are kind of related to the admin giving access to users, but it could be a totally different issue/story entirely:

1) Will users from different teams exist in different orgs/groups/depts within the CMS?

2) What types of content will users from different teams be able to see? Will all users be able to see everything?

If these are valid and you would like me to create UX flow stories for them, I will be more than happy to.

joshkimmell:
@mkiang3051 
will create a research story around roles and permissioning

joshkimmell:
This story is specific to the ability to edit users

PR #572, opened 2016-11-17
Feature/setup new user

### LINKED ISSUE #317: First time setup, path to create an admin user

scottnath:
**NOTE:** this PR is blocked until merging of #560 

Solves the issue of first-time-setup not having users in DB.

---
Resolves #317 


config/default.js
```diff
@@ -124,6 +124,11 @@ module.exports = {
       save: 'save',
       update: 'update',
     },
+    setup: {
+      path: '/create-admin',
```
Snugug: I'm not a fan of having this set in settings. This should be internal to Punchcard. The actions we do this with for internationalization, and that'll be going away once we have i18n in place. Please change this to be a Punchcard internal URL. Too much can go wrong with allowing a user to set the path directly.

lib/users/routes.js
```diff
@@ -261,9 +312,46 @@ const save = (req, res, next) => {
...
+ * @param {object} res - HTTP Response
+ * @param {object} next - Express callback
+ *
+ * @returns {promise} database promise
+ *
+ */
+const first = (req, res, next) => {
+  return database.select('*').from('users').then(rows => {
+    // if any users in DB, we give them a shameful error
+    if (rows.length > 1) {
```
Snugug: Shouldn't this never happen? I'd prefer to see some access control on the first time user creation routes that only allow access if the number of users in a database is greater than zero, as opposed to keeping that page accessible.

views/users/add.html
```diff
@@ -5,12 +5,16 @@
...
+    {% include "_header-not-auth.html" %}
+  {% else %}
+    {{ super() }}
+  {% endif %}
 {% endblock %}
 
 {% block main %}
   <div class="introduction">
-    <h1 class="introduction--h1">{{config.actions.add|capitalize}}</h1>
+    <h1 class="introduction--h1">{{title|capitalize}}</h1>
```
Snugug: This filter is transforming `CMS` in to `cms`. While it may be needed for some generic use cases, it's causing weirdness with this case

lib/users/routes.js
scottnath: I think I'm confused? Or perhaps you are?

`first` is for saving the first user (or superadmin as we may do). It's really just a wrapper for `save` with this db check before triggering `save`. Here in case someone tries to `post` to this url (`/create-admin`) when they shouldn't. 

i.e. this db check _is_ the access control you want, isn't it?

scottnath: adjusted the jsdoc for clarity

config/default.js
scottnath: I agree, however I believe we need some settings which maybe aren't able to be changed by an implementor, but that are still settings. For instance, this path (`/create-admin`) is in a bunch of places in four different files:

```
lib/init/authenticated.js
17:  ... req.url !== config.authentication.login.path ...

lib/routes/login.js
32:          return res.redirect(config.users.setup.path);

lib/routes/users.js
28:    app.get(config.users.setup.path, users.routes.setup);
33:    app.post(config.users.setup.path, multipartMiddleware, users.routes.first);
38:    app.get(`/${config.users.setup.path}/${config.users.actions.save}`, users.routes.setup);

lib/users/routes.js
43:          action: config.users.setup.path,
233:  if (_.includes(referrer, config.users.setup.path)) {
```

It feels wrong to simply hard-code something like a URL path throughout the system. Because if we ever do want to change it, we're on a hunt to find it in all the different places. Same could be said for `config.authentication.login.path` or any number of other system settings that we don't _really_ want to change, but also don't want to keep track of the places they exist in case we ever do.

Regardless, to get this PR complete, I will make this change. But, I think we should consider some internal conglomeration of settings for things like this.

scottnath:
@Snugug changes in. Answering q's:

Questions you're left with:

## `Should we filter the list of roles for first-time users to roles that allow access to /users?`

yes, we _should_, however we do not have roles properly fleshed out to do that. Right now roles/access is a bit raw and we don't have a clear path for access to which pieces. We have [a story for access for roles other than admin](https://github.com/punchcard-cms/punchcard/issues/163) and some general role/access questions that need to be addressed, I believe, before we can say which roles to limit this first-user to. We should also make a decision on `superuser` first.

## `Should this first account work the same way Wordpress/Drupal Superuser accounts work`

I would prefer to have `superuser` be a role in and of itself that can be granted to more than one user. It's nice to be able to delegate sometimes! We could have checks in place for `You cannot delete the last superuser! You must give another account superuser powers before you delete this one`.

bad-person

I don't like the way Wordpress does this anyway. What if the original superuser is leaving a project? Now that specific user account needs to be passed to another person. Emails/passwords changed. Ick!

## `... once the first-time admin account has been created, ... log them in directly?`

Maybe this is a UX question? I feel that it's better to make them sign in so that they reiterate the credentials they just made up. I'm open to either though, but perhaps a separate pr or story for this?

## `..."Add new user" ... use the Password Required form, not the Password Edit form`
bug. fixed in pr.

## `...throwing a validation error... email address that already exists.`

separate PR and story please

scottnath:
No duplicate emails: #574

Snugug:
The only thing I'd very much like to see before this gets merged is automatically logging the user in once they've created an account. Other than that looks good

PR #589, opened 2016-11-29
New component for content types

### LINKED ISSUE #249: Content Creator: Understand content types and take action
_As a_ Content Creator, _I want_ to understand which content type I need and have the ability to take action quickly _so that_ I can create and edit content with ease.

``` gherkin
Feature: Content type list context
  As a Content Creator,
  I want to understand which content type I need
    And have the ability to take action quickly
  So that I can create and edit content with ease.

  Scenario: Understand content types
    Given I am looking to create/edit content in the CMS
     When I view the list of content types
     Then I will see a description for each content type
       And be able to understand the purpose of each content type

  Scenario: Add content
    Given I am looking to create new content
     When I view the list of content types
     Then I will have the ability to add content from this view

  Scenario: View content
    Given I am looking to see all the content for a content type
     When I view the list of content types
     Then I will have the ability to view a list of all the content from a specific content type
```

---

**Content type list context:**
- [ ] Understand content types
- [ ] Add content
- [ ] View content

---

screen shot 2016-08-11 at 2 21 39 pm

http://watson-cms.stage1.mybluemix.net/content

ineedsubstance:
A new component to display content types that include the title, description, and both `Add` and `View` action links.

---
Resolves #249


ineedsubstance:
Waiting on results of this PR #571 in regards to adding the `+` icon to the `Add` button.

ineedsubstance:
Since PR #571 is unresolved, for now, I am removing the `Add` button embellishment since it is not critical to completing this story.

@Snugug - this is ready for your review.

src/sass/partials/components/_content-card.scss
```diff
@@ -0,0 +1,30 @@
+.card {
```
scottnath: Is `.card` generic on purpose? Would this styling be _only_ for the content card, or any card?

lib/routes/data/styles.js
```diff
@@ -63,6 +63,20 @@ module.exports = {
       ],
     ],
   },
+  contentData: {
+    name: 'Careers',
+    description: 'A section realted to joining Watson and what the culture of Watson is about',
```
scottnath: Just for clarity sake, can you make this description sound more like it's describing a content type instead of a section of a website? Slight wording change should prol be enough

ineedsubstance: Hmm... I pulled that description directly from the Careers content type. But for clarity, I can update here.

src/sass/partials/components/_content-card.scss
ineedsubstance: I think it could be for any card, although I'm not sure what those cards would be at this point. @Snugug can comment if this is the right approach or if it should be more specific.

Snugug: If the file name is `content-card` this component should be `.content-card`, not `.card`. Card is a little _too_ generic, maybe we want a more descriptive name, like `basic-card` or `minimal-card`

src/sass/partials/components/_content-card.scss
```diff
@@ -0,0 +1,30 @@
...
+  border: setting-get('input border width') solid setting-get('secondary color');
+  display: flex;
+  flex-direction: column;
+  margin-top: (setting-get('rhythm') / .5);
+  padding: setting-get('rhythm') / .75;
+
+  &--link {
+    @include link(setting-get('active link color'), setting-get('active link color'), setting-get('secondary link color'), setting-get('secondary link color'));
```
Snugug: This is identical to `.base--link`. Instead of duplicating this, add a `base-link` mixin to https://github.com/punchcard-cms/punchcard/blob/master/src/sass/partials/global/_mixins.scss with these defaults (or, alternatively, update the `link` mixin with these as the default arguments) and make sure that `.base--link` calls the mixin

src/sass/partials/components/_content-card.scss
```diff
@@ -0,0 +1,30 @@
...
+    margin-top: setting-get('rhythm') / .5;
+    width: 100%;
+    @include breakpoint(setting-get('responsive form')) {
+      flex-grow: 1;
+      order: 2;
+    }
+  }
+
+  &--button {
+    @include button(setting-get('primary accent color'), setting-get('primary accent color'), setting-get('primary color'), setting-get('primary accent color'), setting-get('primary color'), setting-get('primary color'), setting-get('primary color'));
```
Snugug: Same comment as link above

src/sass/partials/layouts/_content-card.scss
```diff
@@ -0,0 +1,35 @@
+.card {
```
Snugug: * Layouts begin with an underscore, so `._card`
* This, again, needs a more descriptive component name, and the component name and the file name need to match

src/sass/partials/layouts/_content-card.scss
```diff
@@ -0,0 +1,35 @@
...
+  display: flex;
+  flex-direction: column;
+  margin: 2rem 0 0;
+  width: 100%;
+
+
+  @include breakpoint(setting-get('two grid cards')) {
+    margin-right: 4%;
```
Snugug: Where does this 4% come from? That's not in line with _any_ of our rhythm anywhere. We've pretty consistently been doing spacing using `rem` and fixed sizes. Why the sudden change? And why this percentage?

src/sass/partials/layouts/_content-card.scss
```diff
@@ -0,0 +1,35 @@
...
+    width: 48%;
```
Snugug: Where does this 48% come from? Is this suppose to be 50% with a 2% margin?. Width should be 50% and there should be padding in the inside at a fixed width for the gutter

src/sass/partials/layouts/_content-card.scss
```diff
@@ -0,0 +1,35 @@
...
+
+    &:nth-child(2n) {
+      margin-right: 0;
+    }
+  }
+
+  @include breakpoint(setting-get('fixed container')) {
+    margin-right: 4%;
+    width: 22%;
```
Snugug: Again, as above, I'm assuming this is… 25%? With a 3% margin on each side? I'm not sure that math actually works out given the numbers here, but either way, previous comment

views/components/card.html
```diff
@@ -0,0 +1,12 @@
+{% macro card(data) %}
+<article class="card" itemscope itemtype="http://schema.org/Product">
```
Snugug: The items in here don't always semantically contain a product. In fact, the sample data you have here specifically isn't a product. This doesn't necessarily need any more specific type of schema other than `thing` if we can't guarantee the content.

views/styles.html
```diff
@@ -340,5 +340,16 @@ <h2 class="_style-guide--title">Tabular Data</h2>
     {{table(tableData)}}
   </section>
 
+  <section class="_style-guide--group">
+    <h2 class="_style-guide--title">All Content Types</h2>
+    {% from "components/card.html" import card %}
+    <div class="card--group">
```
Snugug: We need a sample with only a single card, and we need a separate template (`views/layouts/card`) that is for the layout. They should be displayed separately in the style guide.

Snugug: Layout needs to match the file name (and therefore the layout name) in the styling

views/components/card.html
```diff
@@ -0,0 +1,12 @@
+{% macro card(data) %}
```
Snugug: Needs to ultimately match the component name above

src/sass/partials/layouts/_content-card.scss
ineedsubstance: ok

src/sass/partials/components/_content-card.scss
ineedsubstance: Just realized that I'm not actually using this class for this component, but I will still make the `base-link` mixin change.

src/sass/partials/layouts/_content-card.scss
ineedsubstance: No sudden change, I still new to the base level styling that you guys have done in the past. No worries, I will adjust to make it consistent.

src/sass/partials/layouts/_content-card.scss
ineedsubstance: Updated to `padding` instead of `margin` and mirrored the `rhythm` to what we are already using on the `form--field-container`. With that said, 48% is used since we have a border and need there to be gutter between the cards.

src/sass/partials/layouts/_content-card.scss
ineedsubstance: updated with same changes as above

views/components/card.html
ineedsubstance: ok

views/styles.html
ineedsubstance: Created the `layouts` stuff, but I'm not sure I'm calling it right in `styles.html`. I gave it a try and you can tell me if it's wrong.

ineedsubstance:
@Snugug - got all your changes in, but wasn't sure about the `layouts` implementation. Anyways, ready for review again. Thanks.

views/layout/basic-card.html
```diff
@@ -0,0 +1,7 @@
+{% from "components/basic-card.html" import basicCard %}
```
Snugug: Until [embed](https://github.com/mozilla/nunjucks/issues/790) lands, the best way to create a sub-layout is to create an include, like you are, but feed data in to it via `{% set %}` before it gets called, and have that be used generate the content, looping over the set content.

Here, you're hard-coding this layout to only use the Style Guide's sample data, which we don't want to do.

(Twig's [embed](http://twig.sensiolabs.org/doc/tags/embed.html) function, for reference)

views/components/basic-card.html
```diff
@@ -0,0 +1,12 @@
+{% macro basicCard(data) %}
+<article class="_basic-card basic-card" itemscope itemtype="http://schema.org/Thing">
```
ineedsubstance: ok

views/layout/basic-card.html
ineedsubstance: I gave this a shot. The routes are not setup yet, so I wasn't sure what exactly to do here. I looked at a portion of the `products` `view` that used `{% set %}` as a [reference](https://github.ibm.com/Watson/wox-ui/blob/develop/views/products/index.html#L12), but it breaks locally so I'm sure it's not right.

src/sass/partials/components/_basic-card.scss
```diff
@@ -1,5 +1,6 @@
 .basic-card {
   border: setting-get('input border width') solid setting-get('secondary color');
+  padding: setting-get('rhythm') / .5;
```
Snugug: * As we've discussed, don't divide by a decimal when you should be multiplying instead. This should be multiplied by 2
* Do we not have a double rhythm setting already? If not, we should

views/layouts/basic-card.html
```diff
@@ -1,7 +1,15 @@
 {% from "components/basic-card.html" import basicCard %}
```
Snugug: Let's just remove this file for now.

src/sass/partials/components/_basic-card.scss
ineedsubstance: Ah, ok. We don't have `double rhythm` set, but I will add it.

views/layouts/basic-card.html
ineedsubstance: OK. I will write a layout story to tackle this then.

ineedsubstance:
@Snugug - is this ready to merge or are there further changes?

Snugug:
Was waiting on the update to pass. Waiting on that again. Will merge once passed

Issue #577, opened 2016-11-21
Core Developer: Breadcrumb Routing


ineedsubstance:
LONG_DESCRIPTION

```gherkin
Feature: FEATURE
  As a PERSONA
  I want NEED
  So that RATIONALE

  Scenario: SCENARIO
    Given CONTEXT
     When ACTION
     Then RESULT
```

---

**FEATURE**
- [ ] SCENARIO

PR #600, opened 2016-11-30
Breadcrumbs component

### LINKED ISSUE #399: Content Creator: Breadcrumbs component
Design and build a component for breadcrumbs navigation.

``` gherkin
Feature: Identify where I am and navigate with ease
  As a Content Creator,
  I want to be able to identify where I am and where I've been in context to the site
  So that I can easily navigate the CMS and so I can retrace my steps if I need to go back.

  Scenario: Breadcrumbs
    Given that, I've navigated past the homepage
      And I see where I am in the website hierarchy
     When I want to navigate up my current page's website hierarchy
     Then I will see the parent hierarchy from my current page
      And each page within the hierarchy will include a title
      And each title will include a link to that respective page
      And the title for the current page will not be linked
      And the title for the current page will have a visual change to indicate to the user that it is not a link
```

---

**Identify where I am and navigate with ease:**
- [ ] Breadcrumbs

---

Reference #338

ineedsubstance:
Created a component for breadcrumbs.

---
Resolves #399 


ineedsubstance:
@poofichu - can you take a look at the visual styling of this and provide feedback, please.

screen shot 2016-11-30 at 4 08 25 pm

poofichu:
This looks FANTASTIC!!!  👍

ineedsubstance:
Great thanks!

@Snugug - this is ready for review please.

lib/routes/data/styles.js
```diff
@@ -63,4 +63,21 @@ module.exports = {
       ],
     ],
   },
+  breadcrumb: {
```
scottnath: This can just be an array called `breadcrumbs` (no need for `item` sub-array), then below you could just use `{% for item in breadcrumbs %}`

This might cause an issue with your macro's name being `breadcrumbs`, but it might not...

lib/routes/data/styles.js
```diff
@@ -63,4 +63,21 @@ module.exports = {
...
+    item: [
+      {
+        title: 'Dashboard',
```
scottnath: Not sure if it's for this PR, but we would need a way to get these titles. Perhaps set up navigation in `config`? Not sure...needs discussion.

lib/routes/data/styles.js
```diff
@@ -63,4 +63,21 @@ module.exports = {
       ],
     ],
   },
+  breadcrumb: {
+    item: [
```
Snugug: Agreed with Scott. As there is no secondary data in breadcrumbs, the actual data model should be an array, not an object with one key that's an array.

src/sass/partials/components/_nav.scss
```diff
@@ -41,3 +41,48 @@
...
   }
 }
+
+.breadcrumb {
+
+  &--list {
+    background-color: transparent;
+    display: flex;
+    margin-top: 0;
+    transform-origin: top center;
```
Snugug: You're not using any CSS transforms in here; why did you set the transform origin?

lib/routes/data/styles.js
Snugug: This is sample data for the breadcrumbs. Where that data is coming from for actual implementation is out of scope for this PR

src/sass/partials/components/_nav.scss
ineedsubstance: My bad, removing.

lib/routes/data/styles.js
ineedsubstance: Got it, will update.

lib/routes/data/styles.js
ineedsubstance: @scottnath - I had created this story for the data #577. Let me know if this is consumable.

lib/routes/data/styles.js
ineedsubstance: I think I got this sorted. Will push my changes.

views/components/breadcrumbs.html
```diff
@@ -1,9 +1,9 @@
-{% macro breadcrumbs(breadcrumb) %}
+{% macro breadcrumbs(data) %}
 <nav itemscope itemtype="http://schema.org/SiteNavigationElement" role="navigation" aria-label="Sub Level Navigation" class="breadcrumb">
   <ul class="breadcrumb--list">
-    {% for item in breadcrumb.item %}
-      <li class="breadcrumb--item" {% if item.active %} data-state="active" {% endif %}>
-        <a href="{{item.url}}" class="breadcrumb--link">{{item.title}}</a>
+    {% for items in data %}
```
scottnath: should stay as singular `item`

ineedsubstance: updated

scottnath:
@Snugug pr awaits your pleasure

src/sass/partials/components/_nav.scss
```diff
@@ -41,3 +41,47 @@
     }
   }
 }
+
+.breadcrumb {
```
Snugug: This should sit in a separate file, `components/_breadcrumb`. While it is a form of navigation, each component file (`_base` excluded) should contain only one file, and it should be the same name as the component inside

src/sass/partials/components/_nav.scss
```diff
@@ -41,3 +41,47 @@
...
+
+  &--list {
+    background-color: transparent;
+    display: flex;
+    margin-top: 0;
+    transition: .15s transform ease-out;
```
Snugug: The transitions [we already have](https://github.com/punchcard-cms/punchcard/blob/master/src/sass/partials/global/_variables.scss) are all pegged at `.2s`. Why is this `.15s`?

src/sass/partials/components/_nav.scss
ineedsubstance: ok got it

src/sass/partials/components/_nav.scss
ineedsubstance: Your right, my mistake. I thought I saw that somewhere but looks like I'm wrong.

src/sass/partials/components/_breadcrumb.scss
```diff
@@ -0,0 +1,43 @@
+.breadcrumb {
```
Snugug: One last nitpick now that I'm reading this over one last time; the component HTML is `breadcrumbs` and the styling is `breadcrumb`. These should be consistent

ineedsubstance:
@Snugug - all changes made. ready for review.

PR #609, opened 2016-12-08
Alert component

### LINKED ISSUE #601: Core Developer: Alert component
Build a component for alert messages.

``` gherkin
Feature: Contextual Feedback
  As a Core Developer,
  I want to provide users with a feedback message that requires their attention 
  So that users understand if an action is needed and what they need to do

  Scenario: Alerts
    Given that, a user may require contextual feedback
    When they perform an action in Punchcard that required feedback
    Then they will see an alert message
      And the message will be displayed in context of the related element 
      And the message will state the issue and how the user can take action
```
---

**Contextual Feedback:**
- [ ] Alerts

---
Related to #225

ineedsubstance:
Created a component for alerts.

---
Resolves #601 

### DIFF
--- a/lib/routes/data/styles.js
+++ b/lib/routes/data/styles.js
@@ -63,4 +63,16 @@ module.exports = {
       ],
     ],
   },
+  alertData: [
+    {
+      for: '00eef724-7cbd-47f5-a735-630510f39711',
+      type: 'Warning!',
+      value: 'This field cannot be left blank!',
+    },
+    {
+      for: 'Login Error',
+      type: 'Oops!',
+      value: 'The email or password you entered is incorrect. Please try again.',
+    },
+  ],
 };
--- a/lib/routes/data/styles.js
+++ b/lib/routes/data/styles.js
@@ -63,10 +63,8 @@ module.exports = {
       ],
     ],
   },
-  alertData: [
-    {
-      for: '00eef724-7cbd-47f5-a735-630510f39711',
-      value: 'Oops! The email or password you entered is incorrect. Please try again.',
-    },
-  ],
+  alertData: {
+    for: '00eef724-7cbd-47f5-a735-630510f39711',
+    value: 'Oops! The email or password you entered is incorrect. Please try again.',
+  },
 };
--- a/lib/routes/data/styles.js
+++ b/lib/routes/data/styles.js
@@ -66,13 +66,7 @@ module.exports = {
   alertData: [
     {
       for: '00eef724-7cbd-47f5-a735-630510f39711',
-      type: 'Warning!',
-      value: 'This field cannot be left blank!',
-    },
-    {
-      for: 'Login Error',
-      type: 'Oops!',
-      value: 'The email or password you entered is incorrect. Please try again.',
+      value: 'Oops! The email or password you entered is incorrect. Please try again.',
     },
   ],
 };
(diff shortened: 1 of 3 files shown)


views/components/alert.html
```diff
@@ -1,3 +1,5 @@
-{% macro alert(for, type, value) %}
-<p class="alert--{{type | upper}}" role="alert" for="{{for}}">{{value}}</p>
+{% macro alert(data) %}
+{% for item in data %}
+<p class="form--alert" role="alert" for="{{item.for}}"><strong class="form--alert-type">{{item.type}} </strong>{{item.value}}</p>
```
scottnath: What criteria determines when `strong` is used?

lib/routes/data/styles.js
```diff
@@ -63,4 +63,16 @@ module.exports = {
       ],
     ],
   },
+  alertData: [
+    {
+      for: '00eef724-7cbd-47f5-a735-630510f39711',
+      type: 'Warning!',
```
scottnath: The alert for forms does not currently have two options. It is just a single message. I think this data should just be `for` and `value`.

views/components/alert.html
scottnath: if we remove `type` from the alert data, can we then get rid of `strong`?

views/components/alert.html
```diff
@@ -1,3 +1,5 @@
-{% macro alert(for, type, value) %}
-<p class="alert--{{type | upper}}" role="alert" for="{{for}}">{{value}}</p>
+{% macro alert(data) %}
+{% for item in data %}
```
scottnath: This component shouldn't check for an array. It should take a single object and create a single alert. If we want multiple components in our styles view, then we should just call the component twice.

ineedsubstance: ok, makes senes.

views/components/alert.html
ineedsubstance: Currently, all alerts look the same and have the same level of visual importance. I was attempting to add some distinction between different alerts by adding the `type`. I will remove for now and create an issue to explore this idea in the future.

screen shot 2016-12-08 at 3 52 53 pm

ineedsubstance: yes, this is related to you comment above.

lib/routes/data/styles.js
ineedsubstance: yes, this is related to you comment above.

ineedsubstance:
@scottnath - changes have been addressed.

views/components/alert.html
scottnath: Still checking for an an array, should not be.

lib/routes/data/styles.js
```diff
@@ -63,4 +63,10 @@ module.exports = {
       ],
     ],
   },
+  alertData: [
```
scottnath: should not be an array, or if it is, each object should be called separately below.

views/styles.html
```diff
@@ -329,6 +329,12 @@ <h1 class="base--h2">Form elements</h1>
   </section>
 
   <section class="_style-guide--group">
+    <h2 class="_style-guide--title">Alerts</h2>
+    {% from "components/alert.html" import alert %}
+    {{alert(alertData)}}
```
scottnath: if you're going to have the data be an array for this, you need your `for` statement here, not in the component

views/components/alert.html
ineedsubstance: ah, my bad. will fix.

lib/routes/data/styles.js
ineedsubstance: yep, got lazy. fixing

views/styles.html
ineedsubstance: nope, fixed above

Issue #588, opened 2016-11-29
Save and Retrieve Files


Snugug:
Given that [Pluggable Storage exists](https://github.com/punchcard-cms/punchcard/issues/492), it now needs to be connected to the form

```gherkin
Feature: Files in Forms
  As a Content Creator
  I want to add a file to my content and have it saved
  So that I can have that file available with my content

  Scenario: Save
    Given that I have uploaded one or more files as part of a piece of content
      And the file uploaded meets all validation criteria
     When I save that content revision
     Then the files should be saved using the configured pluggable storage
      And those files should include its file type, original file name, and new, relative file name

  Scenario: Edit
    Given that I have a content revision that includes files
     When I edit that revision
     Then the revision's files should be viewable to me in the form
      And I should be able to edit, add (where appropriate), and remove (where appropriate) the files
      And any files that I have not edited should be available in the revision being edited when saved
```

---

**Files in Forms**
- [x] Save
- [ ] Edit

scottnath:
@Snugug looks like the validation on input-plugin-file is not ready to validate against actual file data that would happen in a real scenario. Added a [bug report in input-plugin-file](https://github.com/punchcard-cms/input-plugin-file/issues/8) and feel said bug should be cleared before I can move forward on the validation scenario.

scottnath:
moved validating of files to a separate story: https://github.com/punchcard-cms/punchcard/issues/603

PR #610, opened 2016-12-08
Hotfix/session referrer

### LINKED ISSUE #608: Bug: add/edit share session data/referrer
Bug causes `edit` form data to sometimes be in `add` form and vice-versa. Referrer also gets confused. This issue was solved during building of both users and applications section. Need to upgrade referrer and session in content route.

scottnath:
Fixes some old code in content/workflow section. This is required to move forward on #588 

---
Resolves #608 


lib/routes/content.js
```diff
@@ -187,27 +188,26 @@ const routes = application => {
      * @param {object} next - Express callback
     */
     app.get(`/${config.content.base}/:type/:id/:revision/${config.content.actions.edit}`, (req, res, next) => {
-      const errors = _.get(req.session, `form.content.add[${req.params.type.toLowerCase()}].errors`, {});
-      const idSess = _.get(req.session, `form.content.add[${req.params.type.toLowerCase()}].id`, {});
-      const revisionSess = _.get(req.session, `form.content.add[${req.params.type.toLowerCase()}].revision`, {});
-      const values = _.get(req.session, `form.content.add[${req.params.type.toLowerCase()}].content`, {});
+      const errors = _.get(req.session, `form.content.edit[${req.params.type.toLowerCase()}].errors`, {});
+      const id = (_.get(req.session, `form.content.edit[${req.params.type.toLowerCase()}].id`) || _.get(req, 'params.id'));
```
Snugug: Two questions here, for both 192 and 193:
* Why are you using an or statement `||` instead of a default fallback for `_.get` like was previously used?
* Where are these values used? I can't seem to find another use for them, but I do see that this is duplicated in a few places (`id` specifically)

lib/routes/content.js
```diff
@@ -284,18 +284,38 @@ const routes = application => {
...
       }
 
+      // determine data source from referrer
+      if (_.includes(referrer, '/add')) {
+        source.type = 'add';
+      }
+      else if (_.includes(referrer, '/edit')) {
+        source = {
+          type: 'edit',
+          id: _.get(req.session, `form.content.edit[${req.params.type.toLowerCase()}].id`, ''),
```
Snugug: Is this a duplicate of 192?

lib/routes/content.js
scottnath: 1. default fallback for both was an object, which was causing errors. `revision` and `id` must be strings
2. this selection chooses to _either_ grab the existing revision id from the session (in the case of a validation/client error) _or_ grabs the actual revision id from the url

Values used below, `_.set(req.session, 'form.content.edit', {` to re-set the session vars.

These are _only_ used in the case of `values` not being empty - ie after an error. Otherwise, we grab them from the url. (starting line 236)

lib/routes/content.js
scottnath: no. this has different needs. ID can only come from the session because the `save` url does not have any variables in it. Value is doublechecked below (line 333)

lib/routes/content.js
```diff
@@ -284,18 +284,38 @@ const routes = application => {
...
       if (req.body.submit === config.content.actions.new) {
         check = 'save';
       }
 
+      // determine data source from referrer
+      if (_.includes(referrer, '/add')) {
+        source.type = 'add';
+      }
+      else if (_.includes(referrer, '/edit')) {
+        source = {
```
scottnath: btw...`source` object is used in a few spots below. _Mainly_ here to clearly differentiate between adding new / creating a revision.

PR #625, opened 2016-12-19
Feature/filepath

scottnath:
* adds new functions needed for files in content
* removes old content code, which was replaced by content middleware
* adds fixtures for files
* upgrades tests

---
relates to #588 


lib/content/utils.js
```diff
@@ -1,74 +1,95 @@
...
+
+  if (Array.isArray(files) && files.length > 0) {
+    files.forEach(filer => {
+      // if repeatable
+      if (Array.isArray(content[filer.attr][filer.input])) {
+        content[filer.attr][filer.input] = content[filer.attr][filer.input].map(atr => {
+          const attr = atr;
+
+          if (attr.value.hasOwnProperty('original') && attr.value.original !== '') {
+            attr.value.absolute = path.join(config.storage.public, attr.value.relative);
```
Snugug: * Why isn't this using [`storage.get`](https://github.com/punchcard-cms/punchcard/blob/master/lib/storage/get.js)
* Did you test this with HTTP or FTP urls? Things that should look like `method://path`? I'm fairly sure `path.join` destroys the double slash, which is required. That's why `storage.put` and `storage.get` work to normalize that so they can just be concatenated together.

lib/content/utils.js
```diff
@@ -1,74 +1,95 @@
...
+          }
+
+          return attr;
+        });
+      }
+      else {
+        // non-repeatables
+        if (content[filer.attr] && content[filer.attr][filer.input].value.hasOwnProperty('original') && content[filer.attr][filer.input].value.original !== '') {
+          content[filer.attr][filer.input].value.absolute = path.join(config.storage.public, content[filer.attr][filer.input].value.relative);
```
Snugug: Why isn't this using `storage.get`?

Snugug: I don't see any tests to make sure that the actual content being spit out of here is actually correct. No stubs? No fixtures? Need tests for that

tests/apis.js
```diff
@@ -44,7 +44,7 @@ test.serial('Utils: attributes - empty query', t => {
 
   return attributes.then(result => {
     t.is(typeof result, 'object', 'Should contain result, an object.');
-    t.is(Object.keys(result).length, 6, 'Should contain six main objects.');
+    t.is(Object.keys(result).length, 9, 'Should contain six main objects.');
```
Snugug: These descriptions are… wrong. Change the description or fix the test please

tests/content.js
```diff
@@ -1,110 +1,86 @@
...
+  const result = utils.fileinputs(type.attributes);
 
-test('Check content id-not just a number', t => {
-  const rq = (JSON.parse(JSON.stringify(req)));
-  rq.params.id = 123;
-  const result = utils.check.id(rq);
-  t.false(result, 'Not in uuid');
-});
+  t.true(Array.isArray(result), 'should return an array');
+  t.is(result.length, 4, 'should have two items in array');
```
Snugug: Two and four are not the same number

tests/content.js
```diff
@@ -1,110 +1,86 @@
...
+  const type = futils.type('fileinputs');
+  const fileinputs = utils.fileinputs(type.attributes);
+  const result = utils.filepaths(fileinputs, futils.values('fileinputs'));
+
+  t.is(typeof result['fileinputs-file-single'], 'object', 'file-single input should be an object');
+  t.is(typeof result['fileinputs-file-single'].filesingle, 'object', 'file-single input should contain an input object');
+  t.is(typeof result['fileinputs-file-single'].filesingle.value, 'object', 'file-single input should contain an input value object');
+  t.is(typeof result['fileinputs-file-single'].filesingle.value.path, 'string', 'file-single input should contain an absolute string');
+  t.true(_.includes(result['fileinputs-file-single'].filesingle.value.path, config.storage.public), 'file-single input should contain the storage prefix');
+
```
Snugug: I'd add some tests here to ensure that type and relative are in here as well, so it's not _just_ the public URL (to ensure that what's getting passed upwards has those)

tests/fixtures/files/index.js
```diff
@@ -14,42 +13,6 @@ const all = [
 ];
 
 /*
- * Raw File Fixture
- *
- * @returns {array} - An array file objects as one receives from a multi-upload form
- */
-const raw = () => {
```
Snugug: Where did these fixtures disappear to?

scottnath: These are now in *shared tests*. Added [in this pr](https://github.com/punchcard-cms/shared-tests/pull/21)

scottnath: fwiw, I think that `saved` should move there too.

PR #639, opened 2017-01-20
Hotfix/put upload

### LINKED ISSUE #638: Bug: 4 bugs related to file storage
In testing out using Punchcard on bluemix with AWS, I found four bugs.

## Bug 1: value not saved when adding content

When adding a new piece of content, the system was error-failing when trying to grab existing data for the new piece of content. Specifically, knex is creating an `integer` error because the revision variable was`' '` instead of an integer.

## Bug 2: create tmp dir; move middleware

A temp directory needs to be created in the `/public` directory, _after_ it has been created. This requires a slight change to the `upload` middleware, so it should be moved to the init related to routes (should have been put their originally)

## Bug 3: missing local public file dir on fs storage

Same problem as above (directory not being created), except specific to when storing files locally. Should create the `/public/files` directory after the `public` directory is create.

## Bug 4: improper creation of file path

AWS does not like it if the url is not formed correctly. Should use node's `path` to make sure a proper url is created

scottnath:
Fixes four bugs I found. See #638 for bug details.

Note: probably easier to review if you review each commit separately.
---
Resolves #638

### DIFF
--- a/lib/init/routes.js
+++ b/lib/init/routes.js
@@ -28,6 +28,11 @@ module.exports = (app) => {
     // creates temp storage location if it does not exist
     mkdirp.sync(dest);
 
+    // creates final storage location when using file storage
+    if (config.storage.type === 'fs' && typeof config.storage.dest === 'string' && config.storage.dest !== '') {
+      mkdirp.sync(config.storage.dest);
+    }
+
     // add all content types and flows to request object
     app.use((req, response, next) => {
       const request = req;
--- a/lib/init/routes.js
+++ b/lib/init/routes.js
@@ -1,11 +1,33 @@
 'use strict';
 
 const _ = require('lodash');
+const config = require('config');
+const multer = require('multer');
+const mkdirp = require('mkdirp');
+
 const content = require('../content/middleware');
 const users = require('../users/middleware');
 
 module.exports = (app) => {
   return new Promise(res => {
+    // temp file storage location
+    const dest = config.storage.temp.dest || 'public/tmp/';
+
+    // multer form handling setup
+    const upload = multer({
+      storage: multer.diskStorage({
+        destination: (req, file, cb) => {
+          cb(null, dest);
+        },
+        filename: (req, file, cb) => {
+          cb(null, `/${file.originalname}`);
+        },
+      }),
+    });
+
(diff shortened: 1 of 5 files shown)


scottnath:
@Snugug checking in to see if you have time to review.

note: moving forward with possible use of punchcard in an internal project - if you cannot review this or others, we need to come up with an alternative review process for PRs in Punchcard

lib/init/routes.js
```diff
@@ -1,11 +1,38 @@
...
+          cb(null, dest);
+        },
+        filename: (req, file, cb) => {
+          cb(null, `/${file.originalname}`);
+        },
+      }),
+    });
+
+    // creates temp storage location if it does not exist
+    mkdirp.sync(dest);
```
Snugug: Why are we using this instead of `fs-extra` which I think it's already a dependency

lib/routes/content.js
```diff
@@ -317,34 +304,31 @@ const routes = application => {
...
       let audits;
       let publishable = false;
       let check = 'publish';
       let files = [];
       let source = {
-        type: '',
+        type: 'add',
         id: '',
-        revision: '',
+        revision: Math.floor(1000000 + Math.random() * 9000000),
```
Snugug: Why a random number instead of a uuid?

Snugug:
Sorry about the delay, forgot you had requested a review. Just ping me if I forget again

lib/init/routes.js
scottnath: fs-extra is [not currently a dependency](https://github.com/punchcard-cms/punchcard/blob/master/package.json#L59), which is why I chose this more targeted module. Can switch if that's preferred.

lib/routes/content.js
scottnath: `revision` content is a number, not a uuid. _Must_ be a number, (as opposed to blank, or string) or we get an integer error from knex.

scottnath:
@Snugug no biggie on delay this time. We should talk though, about going forward and how we want to approach continuing dev work on punchcard.

comments added to your review! thanks for that

scottnath:
@Snugug can I get a merge please?

